PR #15, opened 2021-09-21
Tweaks, cherry-pick and tests

sugarraysam:
Fixing a few things:
- Add more pre-commit (requires splitting pylint target)
- Clean dependencies in setup.py
- added some extra Makefile targets
- change make generate so it uses pipenv
- fix --user inside container
- bump to python version 3.9

Also cherry-picked few extra tests.

Do we want to keep `Pipfile.lock` in git? I added it to `.gitignore` for now as it was causing some issues for me.

@mayankshah1607


mayankshah1607:
@sugarraysam Looks like the README has a conflict. Could you rebase against `main` please? :)

mayankshah1607 (approved):

PR #21, opened 2021-10-06
Make package work with python 3.6

mayankshah1607:

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -21,4 +21,4 @@ yamllint = "1.26.3"
 managedtenants_cli = {editable = true,path = "."}
 
 [requires]
-python_version = "3.9"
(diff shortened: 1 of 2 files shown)


Pipfile
```diff
@@ -21,4 +21,4 @@ yamllint = "1.26.3"
 managedtenants_cli = {editable = true,path = "."}
 
 [requires]
-python_version = "3.9"
+python_version = "3"
```
mayankshah1607, line 24: Currently there doesn't seem to be a way to provide a range of verisons like `">=3.6"`. Setting it to `"3"` will make sure the it works on all python versions, although we should try and avoid verisons lower than `"3.6"`

PR #34, opened 2021-11-04
[MTSRE-295] Add --only-changed flag to modify the addons_loader behavior and speedup pipeline.

sugarraysam:
In an effort to reduce our bundles-to-prod deployment time, adding this piece of code to modify the `addons_loader`.

# TODO after merging

- [ ] Modify `managedtenants` CLI command (and maybe docs) as the `--dry-run` flag now needs to be placed **BEFORE** any subcommand (`load` or `run`)
- [ ] Modify `Makefile` target and add  `--only-changed` flag

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,13 +30,13 @@ repos:
       - id: trailing-whitespace
   # Black
   - repo: https://github.com/psf/black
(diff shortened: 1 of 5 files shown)


mayankshah1607 (approved):
LGTM!

Minor nit: s/childs/children/g

PR #35, opened 2021-11-04
[MTSRE-278] Enforce python3.9 and bump sretoolbox.

sugarraysam:
Also added some cleanup.

This is required as sretoolbox is now Python >= 3.8 and we will need it to utilize the `Mtcli` integration.

Also migrating logger code to sretoolbox.

# TODO (after merge)
- [x] Bump sretoolbox to include Mtcli: https://github.com/app-sre/sretoolbox/pull/62
- [ ] Bump VERSION of managed-tenants-cli + new release
- [ ] Merge this: https://<gitlab>/managed-tenants/-/merge_requests/1456


erdii:
Do we have python 3.9 available in our CI?

sugarraysam:
Using docker-in-docker for now: https://_gitlab_/managed-tenants/-/merge_requests/1456

ashishmax31:
Don't we have to merge gitlab/service/managed-tenants/-/merge_requests/1456 before we release this new version?
Looks good otherwise

ashishmax31 (approved):

yashvardhan-kukreja:
> Don't we have to merge gitlab/service/managed-tenants/-/merge_requests/1456 before we release this new version?

I don't think so. Please correct me if I am wrong but isn't it true that currently, the HEAD of `managed-tenants` isn't even making use of `mtcli`?

And the MR - https://gitlab.cee.redhat.com/service/managed-tenants/-/merge_requests/1456 - is the first-time intro of `mtcli`'s usage in the `managed-tenants`. Hence, wouldn't merging this pull request #35 first make more sense (because it would bring mtcli to the latest suitable state) and then, merging https://gitlab.cee.redhat.com/service/managed-tenants/-/merge_requests/1456 so that that MR is forced and ensured to be compliant with the latest state of `mtcli`?

cc: @sugarraysam

sugarraysam:
@ashishmax31 We need to first release a new version of managed-tenants-cli that includes sretoolbox 1.2.0. This is for two reasons:
1. Want to use `Mtcli` binary abstraction from sretoolbox: https://github.com/app-sre/sretoolbox/blob/master/sretoolbox/binaries/mtcli.py
2. The `managed-tenants` repository inherits it's verion of sretoolbox from importing the `managed-tenants-cli` package. It's a not-so-nice side effect if you ask me: `managed-tenants/-/blob/main/requirements-dev.txt`

So we need to merge this first, and then we will be able to write a custom task using `sretoolbox.binaries.mtcli.Mtcli`. Does this make sense?

PR #36, opened 2021-11-09
[MTSRE-214] Post addon data to versions endpoint

mayankshah1607:
This PR allows the tooling to post addon data to the new versions endpoint in the OCM API [1]. 

- If an Addon uses ImageSets, data is posted to the new versions endpoint as well as the old endpoint. In this case, data posted to the old endpoint will not contain versioned fields such as parameters, requirements, etc.
- If an Addon does not use ImageSets, data is posted only on the old endpoint, hence maintaining backwards compatibility.

Once this MR [2] is merged, versioned fields will be stored along with the bundles in `managed-tenants-bundles` which will be used to populate the ImageSets in `managed-tenants`. These ImageSets in turn will be used for posting data to the new versions endpoint.

[1] https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/merge_requests/3298
[2] https://gitlab.cee.redhat.com/service/managed-tenants-bundles/-/merge_requests/157


managedtenants/utils/ocm.py
```diff
@@ -117,6 +131,11 @@ def disable_addon(self, addon_id):
         )
 
     def upsert_addon(self, metadata):
+        if metadata.get("addonImageSetVersion"):
+            try:
+                addon = self.add_addon_version(metadata)
+            except OCMAPIError as exception:
+                raise exception
```
sugarraysam, line 138: Do we need to fallback to a _patch request if response code is 409? Similar to the `add_addon` case? It is poorly documented (api.openshift.com) but feels the behavior of OCM should be the same as the non-versioned case.

```python
self._patch(f"/api/clusters_mgmt/v1/addons/{addon_id}/versions/{version_id}", json=imageset)
```

Having some OCM integration tests would be quite useful to validate this. It's probably quite involved, but there is a way to run it locally in a container. WDYT?

Suggested refactor (drop extra try/except):

```python
def upsert_addon(self, metadata):
    try:
        if metadata.get("addonImageSetVersion"):
            addon = self.add_addon_version(metadata)
        else:
            addon = self.add_addon(metadata)
    except OCMAPIError as exception:
        if exception.response.status_code == 409:
            if metadata.get("addonImageSetVersion"):
                return self.update_addon_version(metadata)
            else:
                return self.update_addon(metadata)
        raise exception

    return addon
```

managedtenants/utils/ocm.py
```diff
@@ -127,12 +146,31 @@ def upsert_addon(self, metadata):
...
+
     def _addon_from_metadata(self, metadata):
         addon = {}
         metadata["addOnParameters"] = metadata.get("addOnParameters", [])
         metadata["addOnRequirements"] = metadata.get("addOnRequirements", [])
         for key, val in metadata.items():
             if key in self.ADDON_KEYS:
+                # Skip adding these parameters as they're present
+                # in the ImageSet (versions endpoint)
+                if metadata.get("addonImageSetVersion") and key in [
```
sugarraysam, line 168: nit: use set operations:

```python
if metadata.get("addonImageSetVersion") and key in (ADDON_KEYS - IMAGESET_KEYS):
```

sugarraysam (commented):
We need to check if the versions endpoint returns a 409 in case the version already exists (and needs to be patched).

Some OCM integration test would be SOO nice if you can pull it off in this PR -> otherwise please create a ticket :)

managedtenants/utils/ocm.py
mayankshah1607: The intention of this condition is to ensure that the versioned fields are not populated into the addon endpoint payload (old endpoint) if the addon uses imagesets (because these fields will be posted against the new versions endpoint instead, so no need to populate them here anymore).

The only extra field in `IMAGESET_KEYS` is `indexImage` which will anyway always be absent if `addonImageSetVersion` is present. 

We could use `IMAGESET_KEYS` too, shouldn't make a difference. IMO I just wanted to be explicit about these 3 fields as they're the ones that should actually be skipped when using the new endpoint. `indexImage` will simply never be present in the case where we're using imageSets.

managedtenants/utils/ocm.py
mayankshah1607: > Suggested refactor (drop extra try/except):

@sugarraysam  This refactor would allow the addon only to only be posted to either endpoints. What we have to do instead is:
- Post the addon to both endpoints if it uses ImageSets
- Post the addon only to the old endpoint if it used indexImage field

Reference issue - https://issues.redhat.com/browse/MTSRE-214

mayankshah1607: I just refactored this a bit that reduced duplicate code. Let me know what you think :)

sugarraysam: :heart: im a big fan of that refactor. I totally mislead you , please never trust me haha

ashishmax31:
lgtm

ashishmax31 (approved):

sugarraysam:
lets merge this , and work on OCM integration tests in another ticket.

PR #37, opened 2021-11-15
Add managed-tenants bundles package

ashishmax31:
Adds a new package which provides bundle image and index image building/pushing functionalities.


managedtenants/bundles/utils.py
```diff
@@ -0,0 +1,61 @@
+
+from managedtenants.utils.quay_api import QuayApi
+import subprocess
+
+
+def run(cmd, logger=None):
```
sugarraysam, line 6: nit: Should we move this to a general utils file? (also used in `git.py`)

managedtenants/utils/logger.py
```diff
@@ -0,0 +1,27 @@
+import logging
```
sugarraysam, line 1: we can trash this, it has been migrated to sretoolbox

sugarraysam (approved):
Added few comments, main points are:
- We need the 0 size index image check (or every image for that matter, maybe add check before pushing an image)
- We are not creating `quay_repo` as part of the tooling, only checking if they exist, assuming this will be implemented in `ci.py`?
- Last we could have migrated `Gitlab` code as well (`gitlab_client.py`) but this can be another PR.

Thanks for doing this task. Refactoring is never the most exciting. Way to take one for the team :heart:

ashishmax31:
@sugarraysam 
> We need the 0 size index image check (or every image for that matter, maybe add check before pushing an image)

So i tried to implement that functionality here, but in doing that i found out that I'm basically doing the same thing i did in the `mtcli` (bundle extraction from index image) binary(Although doing it in python seemed a little tricky). So i'm thinking to add a new subcommand in mtcli (something like `mtcli list-bundles <indexImage>`) to list the bundles inside an indexImage and use that here. I think it'll be useful for our tenants as well, since opm doesn't have a straight-forward command to list the bundles in an indeximage.

sugarraysam:
I like that idea. Because listing bundles is quite involved (run grpc server inside docker image, query appropriate RPC endpoints, filter for some fields), anything we can do to abstract it in one command would be awesome.

managedtenants/bundles/bundle_utils.py
```diff
@@ -0,0 +1,298 @@
...
+            self.logger = logger
+        else:
+            self.logger = get_text_logger("managedtenants-bundles")
+
+    def validate_local_bundles(self):
+        """
+        Validates the all the bundles present locally under
+        the given addon directory.
+        """
+        if not self.dry_run:
```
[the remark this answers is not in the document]
ashishmax31: Missed this.

ashishmax31:
> * We are not creating `quay_repo` as part of the tooling, only checking if they exist, assuming this will be implemented in `ci.py`?

Yes, this package expose a fn to create a quay repo, the consumers of the package should use that to create the repo before starting to push bundle/index-images.

> * Last we could have migrated `Gitlab` code as well (`gitlab_client.py`) but this can be another PR.
  
 There is already a file called `gitlab_client.py`(https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/utils/gitlab_client.py) which has all the required code for interacting with Gitlab API's.

PR #39, opened 2021-11-15
[MTSRE-303] Allow subscription configs to be created

ashishmax31:
This PR allows tenants to create subscription configs for their addons by specifying them in the addon metadata file. (addon.yaml file) 

Todo after this is merged: 
- Update docs

### DIFF
--- a/docs/tenants/zz_schema_generated.md
+++ b/docs/tenants/zz_schema_generated.md
@@ -223,3 +223,13 @@
     - **`operator_namespace`** *(string)*
 
     - **`enabled`** *(boolean)*
+
+- **`config`** *(object)*: Cannot contain additional properties.
+
+  - **`env`** *(array)*
+
+    - **Items** *(object)*: Cannot contain additional properties.
+
+      - **`name`** *(string)*
+
+      - **`value`** *(string)*
--- a/docs/tenants/zz_schema_generated.md
+++ b/docs/tenants/zz_schema_generated.md
@@ -224,9 +224,9 @@
 
     - **`enabled`** *(boolean)*
 
-- **`config`** *(object)*: Cannot contain additional properties.
+- **`config`** *(object)*: Configs to be passed to the subscription OLM object. Cannot contain additional properties.
 
-  - **`env`** *(array)*
+  - **`env`** *(array)*: Array of environment variables (name, value pair) to be created as part of the subscription object.
 
     - **Items** *(object)*: Cannot contain additional properties.
 
--- a/managedtenants/core/addons_loader/addon.py
+++ b/managedtenants/core/addons_loader/addon.py
@@ -21,7 +21,7 @@ from managedtenants.utils.schema import (
 # These addon IDs _MUST_ be stable and not changed or bad things will happen
 # (this applies to addon IDs in general, but it's worth pointing out here, too)
 _ADDON_OPERATOR_ADDON_IDS = ["reference-addon"]
-_PERMITTED_SUBSCRIPTION_CONFIGS = {"env"}
+_PERMITTED_SUBSCRIPTION_CONFIGS = ["env"]
 
 
 class Addon:
@@ -107,9 +107,11 @@ class Addon:
     def _validate_subscription_config(self, metadata):
         if not metadata.get("config"):
             return
-        configs_present = set(metadata["config"].keys())
+        configs_present = metadata["config"]
         # configs_present should be a subset of _PERMITTED_SUBSCRIPTION_CONFIGS
-        if not configs_present <= _PERMITTED_SUBSCRIPTION_CONFIGS:
+        for item in configs_present:
+            if item in _PERMITTED_SUBSCRIPTION_CONFIGS:
+                continue
             raise AddonLoadError(
                 f"{self.path} validation error: "
                 "Unsupported subscription config objects"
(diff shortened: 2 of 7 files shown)


tests/sss/test_subscription_config.py
```diff
@@ -0,0 +1,32 @@
...
+    expected_values = [
+        {"name": "LOCATION", "value": "Black Mesa Research Facility"},
+        {"name": "USER", "value": "Gordon Freeman"},
+    ]
+    addon = request.getfixturevalue(addon_str)
+    sss_walker = addon.sss.walker()
+    subscription_obj = sss_walker["sss_deploy"]["spec"]["resources"][
+        "Subscription"
+    ][0]
+    if addon.subscription_config_present():
```
mayankshah1607, line 24: Do we actually need this method? I see that its only being used in the test. We can also simply replace it with:
```suggestion
    if addon.metadata.get("config"):
```

docs/tenants/zz_schema_generated.md
```diff
@@ -223,3 +223,13 @@
...
+
+- **`config`** *(object)*: Cannot contain additional properties.
+
+  - **`env`** *(array)*
+
+    - **Items** *(object)*: Cannot contain additional properties.
+
+      - **`name`** *(string)*
+
+      - **`value`** *(string)*
```
mayankshah1607, line 235: Can we make these more descriptive?

mayankshah1607 (changes requested):

tests/sss/test_subscription_config.py
ashishmax31: I think that method can stay, `subscription_config_present` abstracts the implementation detail. I actually forgot to use that in the jinja template, fixed it now.

managedtenants/core/addons_loader/addon.py
```diff
@@ -21,6 +21,7 @@
 # These addon IDs _MUST_ be stable and not changed or bad things will happen
 # (this applies to addon IDs in general, but it's worth pointing out here, too)
 _ADDON_OPERATOR_ADDON_IDS = ["reference-addon"]
+_PERMITTED_SUBSCRIPTION_CONFIGS = {"env"}
```
apahim, line 24: Sets are more expensive than tuples or lists. Unless you need on-the-fly deduplication, I'd reccomend using a tuple here. Or a list, for consistency with the previous line.

managedtenants/core/addons_loader/addon.py
```diff
@@ -93,12 +97,26 @@ def load_metadata(self, environment):
...
 
         if "extraResources" in metadata:
             self.extra_resources_loader = FileSystemLoader(str(metadata_dir))
 
         return metadata
 
+    def _validate_subscription_config(self, metadata):
+        if not metadata.get("config"):
+            return
+        configs_present = set(metadata["config"].keys())
```
apahim, line 110: `metadata["config"] `is a dict, no need to call `keys()`, the dict iterator returns the keys anyway.

apahim: why do you need to deduplicate it? the schema wont allow the same key twice.

managedtenants/core/addons_loader/addon.py
```diff
@@ -93,12 +97,26 @@ def load_metadata(self, environment):
...
+        # configs_present should be a subset of _PERMITTED_SUBSCRIPTION_CONFIGS
+        if not configs_present <= _PERMITTED_SUBSCRIPTION_CONFIGS:
```
apahim, line 112: This is hard to read... I'd suggest `if configs_present not in _PERMITTED_SUBSCRIPTION_CONFIGS`

managedtenants/core/addons_loader/addon.py
ashishmax31: @apahim I just wanted to create a set out of the keys, so that i could check if its a subset of the allowed keys.

managedtenants/core/addons_loader/addon.py
Ajpantuso: Wouldn't the inverse condition read closer to the true intention? (And maybe s/item/config/)
```python
for config in configs_present:
    if config not in _PERMITTED_SUBSCRIPTION_CONFIGS:
        raise AddonLoadError(...)
```
Minor nit so please ignore if you prefer the existing logic.

apahim: I tend to keep conditions un-nested, for readability. Personal choice really, I understand your reasoning and it would be fine as well :)

Ajpantuso: I _really_ prefer ugly one-liners anyway :)
```python
if any(True for x in xs if x not in valid_xs):
```

ashishmax31: And I prefer writing declarative code over imperative :D

apahim: Maybe I read PEPs too much... PEP20 says:

- Flat is better than nested.
- Readability counts.

That's where I come from :)

apahim: Oh, and that topic is covered in the SREs style guide... https://service.pages.redhat.com/dev-guidelines/docs/appsre/python-style-guide/#nested-conditions

feel free to contribute to it, it's a collection of practices we try to follow.

Ajpantuso: Nice! I'll take a look. And yes definitely nested `if`s is a smelly smell indeed.

mayankshah1607 (approved):

PR #41, opened 2021-11-18
MTSRE-279 | Move deadmanssnitch configuration under the metadata schema

yashvardhan-kukreja:

### DIFF
--- a/managedtenants/data/metadata.schema.yaml
+++ b/managedtenants/data/metadata.schema.yaml
@@ -217,21 +217,25 @@ properties:
     required:
     - tags
     properties:
-      clusterDeploymentSelector: ## default: check selectorsyncset.yaml.j2 - line 260
+      clusterDeploymentSelector:
         type: object  ## should we go as granular as describing properties for matchExpressions/matchLabels?
-      snitchNamePostFix: ## default: {{ADDON.metadata['id']}}
+        description: "Default: check selectorsyncset.yaml.j2 - line 260"
+      snitchNamePostFix:
         type: string
         pattern: ^[A-Za-z0-9][A-Za-z0-9-]{0,60}[A-Za-z0-9]$
+        description: "Default: {{ADDON.metadata['id']}}"
       targetSecretRef:
         type: object
         additionalProperties: false
         properties:
-          name: ## default: {{ADDON.metadata['id']}}-deadmanssnitch
+          name:
             type: string
             pattern: ^[A-Za-z0-9][A-Za-z0-9-]{0,60}[A-Za-z0-9]$
-          namespace: ## default: {{ADDON.metadata['targetNamespace']}}
+            description: "Default: {{ADDON.metadata['id']}}-deadmanssnitch"
+          namespace:
             type: string
             pattern: ^[A-Za-z0-9][A-Za-z0-9-]{0,60}[A-Za-z0-9]$
+            description: "Default: {{ADDON.metadata['targetNamespace']}}"
       tags:
         type: array
         items:
--- a/managedtenants/data/metadata.schema.yaml
+++ b/managedtenants/data/metadata.schema.yaml
@@ -210,6 +210,30 @@ properties:
       secretNamespace:
         type: string
         pattern: ^[A-Za-z0-9][A-Za-z0-9-]{0,60}[A-Za-z0-9]$
+  deadmanssnitch:
+    type: object
+    additionalProperties: false
+    properties:
+      clusterDeploymentSelector: ## default: check selectorsyncset.yaml.j2 - line 260
+        type: object  ## should we go as granular as describing properties for matchExpressions/matchLabels?
+      snitchNamePostFix: ## default: {{ADDON.metadata['id']}}
+        type: string
+        pattern: ^[A-Za-z0-9][A-Za-z0-9-]{0,60}[A-Za-z0-9]$
+      targetSecretRef:
+        type: object
+        additionalProperties: false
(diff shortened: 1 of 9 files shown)


yashvardhan-kukreja:
Current status WIP: tests are yet to be fixed

managedtenants/data/metadata.schema.yaml
```diff
@@ -210,6 +210,30 @@ properties:
       secretNamespace:
         type: string
         pattern: ^[A-Za-z0-9][A-Za-z0-9-]{0,60}[A-Za-z0-9]$
+  deadmanssnitch:
```
sugarraysam, line 213: Please add some `description: ` field to help developers and tenants. They will be rendered as markdown as well: https://github.com/mt-sre/managed-tenants-cli/blob/main/docs/tenants/zz_schema_generated.md

managedtenants/data/metadata.schema.yaml
```diff
@@ -210,6 +210,30 @@ properties:
...
+        type: object
+        additionalProperties: false
+        properties:
+          name: ## default: {{ADDON.metadata['id']}}-deadmanssnitch
+            type: string
+            pattern: ^[A-Za-z0-9][A-Za-z0-9-]{0,60}[A-Za-z0-9]$
+          namespace: ## default: {{ADDON.metadata['targetNamespace']}}
+            type: string
+            pattern: ^[A-Za-z0-9][A-Za-z0-9-]{0,60}[A-Za-z0-9]$
+      tags: ## TODO(ykukreja): should the tags be enforced to be required?
```
sugarraysam, line 232: I think we can safely make `tags:` a required field.

managedtenants/data/selectorsyncset.yaml.j2
```diff
@@ -247,6 +247,52 @@ items:
...
+      name: deadmanssnitch-api-key
+      namespace: deadmanssnitch-operator
+    
+    {% if ADDON.metadata['deadmanssnitch']['snitchNamePostFix'] is defined %}
+    snitchNamePostFix: {{ADDON.metadata['deadmanssnitch']['snitchNamePostFix']}}
+    {% else %}
+    snitchNamePostFix: {{ADDON.metadata['id']}} {# Ref: https://gitlab.cee.redhat.com/service/managed-tenants/-/blob/main/addons/ocs-converged-qe/metadata/stage/deadmanssnitch.yaml.j2#L16 and https://gitlab.cee.redhat.com/service/managed-tenants/-/blob/main/addons/ocs-converged/metadata/stage/deadmanssnitch.yaml.j2#L16 #}
+    {% endif %}
+    
+    {% if ADDON.metadata['deadmanssnitch']['tags'] is defined %}
```
sugarraysam, line 278: Let's make `tags:` required and drop the if/else.

sugarraysam (approved):
LGTM except the few comments I left.

@apahim can you please double check? You have a much better knowledge of DeadmansSnitch, and also originally suggested unifying our extraResources interface.

managedtenants/data/metadata.schema.yaml
```diff
@@ -210,6 +210,33 @@ properties:
       secretNamespace:
         type: string
         pattern: ^[A-Za-z0-9][A-Za-z0-9-]{0,60}[A-Za-z0-9]$
+  deadmanssnitch:
+    type: object
+    additionalProperties: false
+    description: "Denotes the Deadmans Snitch Configuration which is supposed to be setup alongside the Addon."
+    required:
+    - tags
+    properties:
```
apahim, line 219: I'm under the impression we are exposing way more than the minimum required. The only properties that seem relevant for the tenants to customize are `tags` and `targetSecretRef`. Everything elase seems to be templated out of the other addon.yaml properties...

yashvardhan-kukreja: Even I felt so but when I looked at the managed-tenants gitlab repository, there were some (or rather many) add-ons whose `deadmansnitch.yaml.j2`   consisted of custom `clusterDeploymentSelector` , `snitchNamePostFix`, `targetSecretRef`, etc. (all fields except dmsApiKeyRef for some or other addon) which didn't comply well with their addon metadata parameters (addon.yaml) the way they were supposed to.

For examples of such add-ons, you can check out:
https://gitlab.cee.redhat.com/service/managed-tenants/-/merge_requests/1507

and take a look at the Addon metadata-s in the MR where I have specified more fields under `deadmanssnitch` field apart from `tags`.

That's why I made the `deadmanssnitch` section tweakable for now so that at least we can proceed to deprecate the `extraResources` and migrate the Addons (having dms configuration) to make use of `deadmanssnitch` field and get rid of `extraResources` field and all the `deadmansnitch.yaml.j2` files.
This gives us the freedom to later on incrementally fix all the addons' deadmanssnitch configuration in such a way that they don't end up relying on custom fields under `deadmanssnitch` section except `tags`.

And from the tenants' perspective, they need not worry about setting all these fields because all of the "tweakable" fields under `deadmanssnitch` section have rightful defaults set. So, they can just set `deadmanssnitch.tags` and rest of the dms configuration would be taken care of.

Let me know your thoughts :)

apahim: looking at those, it makes sense what you did. We should probably cleanup those differences, I believe they have no reason to exist. But let's do one change at a time... we first do as you say, later we can look into talking to our tenants to drop all that custom configuration which is not really needed.

apahim:
lgtm

PR #42, opened 2021-11-22
[draft]Imageset subscription config support

ashishmax31:
Update tooling to support subscription configs in imageset files.

### DIFF
--- a/managedtenants/core/addons_loader/addon.py
+++ b/managedtenants/core/addons_loader/addon.py
@@ -66,7 +66,20 @@ class Addon:
         return self.path.name
 
     def subscription_config_present(self):
-        return self.metadata.get("config") is not None
+        # If imageset is present, always check for subscriptionConfig
+        # in the imageset file.
+        if self.imageset:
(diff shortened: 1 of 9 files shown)


mayankshah1607:
I'm not quite sure why we should be versioning subscription configs. IMO this would just make things slightly complicated as now managed-tenants-bundles and CPaaS would both be responsible for ensuring that subscription configs are populated within the imagesets.

Let's see what others think

ashishmax31:
>I'm not quite sure why we should be versioning subscription configs.
It'd be for the same reason for which we've versioned `addOnParameters `, `subOperators` etc. In that line of logic, i think even subscription_config should be versioned.(Especially since the features supported by SubscriptionConfig are more than just passing env variables.)

ashishmax31:
@yashvardhan-kukreja Your approach could work, although the tenants -> cpaas interface would still have to accommodate this feature (to keep it clean, would be a bad design to ask tenants to come and modify the imageset files created by cpaas.)

ashishmax31:
Will reopen this PR with the mentioned approach

PR #45, opened 2021-11-30
Imageset subscription config support

ashishmax31:
Allow addon imageset's to define subscription configs

### DIFF
--- a/managedtenants/core/addons_loader/addon.py
+++ b/managedtenants/core/addons_loader/addon.py
@@ -113,27 +113,11 @@ class Addon:
 
         self._validate_schema(metadata)
         self._validate_extra_resources(environment, metadata)
-        self._validate_subscription_config(src=metadata)
 
         if "extraResources" in metadata:
             self.extra_resources_loader = FileSystemLoader(str(metadata_dir))
 
         return metadata
-
-    def _validate_subscription_config(self, src):
-        if not src.get("subscriptionConfig"):
-            return
-        configs_present = src.get("subscriptionConfig")
-        # configs_present should be a subset of _PERMITTED_SUBSCRIPTION_CONFIGS
-        for item in configs_present:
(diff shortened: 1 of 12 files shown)


yashvardhan-kukreja:
Awesome work! @ashishmax31. 

Just a minor nit: can you run `make generate` to auto-generate some required changes complying with your PR :)

yashvardhan-kukreja (approved):

sugarraysam:
> Awesome work! @ashishmax31.
> 
> Just a minor nit: can you run `make generate` to auto-generate some required changes complying with your PR :)

@ashish you probably want to install pre-commit hooks:
```bash
$ pre-commit install
```

managedtenants/core/addons_loader/addon.py
```diff
@@ -66,7 +66,23 @@ def name(self):
...
-        return self.metadata.get("config") is not None
+        # If imageset is present, check for subscriptionConfig in the
+        # imageset file, otherwise check for the default subscription config
+        # in the addon metadata file.
+        if self.imageset:
+            if self.imageset.get("subscriptionConfig"):
+                return True
+        return self.metadata.get("subscriptionConfig") is not None
+
+    def get_subscription_config(self):
```
sugarraysam, line 77: nit: do we need the `subscription_config_present` function? it seems in any case, the `get_subscription_config` will return `None` if the config is not there, so we could really just do:

```python
def get_subscription_config(self):
  if self.imageset and self.imageset.get("subscriptionConfig"):
      return self.imageset.get("subscriptionConfig", None)
  return self.metadata.get("subscriptionConfig", None)
```

sugarraysam (approved):
lgtm, good work !

managedtenants/core/addons_loader/addon.py
ashishmax31: Yep, that works

PR #46, opened 2021-12-01
Attempting to fix twine file exist error.

sugarraysam:
Tested on testpypi: https://packaging.python.org/guides/using-testpypi/

```bash
$ python -m twine upload --skip-existing --verbose --repository testpypi dist/*
```


sugarraysam:
The actual issue was VERSION was bumped without a corresponding tag. Closing this.

PR #50, opened 2021-12-02
ocm: Fix addon version patch operation

mayankshah1607:
This PR removed the `id` field from the payload while executing a PATCH operation against the OCM versions API. This would otherwise lead to the following error:

```
original error: {"kind":"Error","id":"400","href":"/api/clusters_mgmt/v1/errors/400","code":"CLUSTERS-MGMT-400","reason":"The following attributes are not allowed: 'id'","operation_id":"49b91421-924c-4cef-a2ed-020f3a34dfa1"}
```

This fix has been tested locally against the OCM API.


ashishmax31 (approved):
lgtm

PR #56, opened 2021-12-06
Parse imageset version from imageset name for OCM upsert

mayankshah1607:
The current approach is to use `metadata.addonImageSetVersion`. However, the problem with this is that the said field could be set to `latest`, which causes issues when posting data to OCM, as `latest` is not recognized as a version. Instead, this PR parses the version directly from the imageset name instead of relying on what is set in the metadata


yashvardhan-kukreja (approved):
lgtm!

PR #60, opened 2021-12-08
remove: __init__ from tests/

yashvardhan-kukreja:


mayankshah1607:
We have non-test code within the tests/ directory that is being imported, so I'm not sure if directly removing the init file like this will work

yashvardhan-kukreja:
yeah I was trying something

PR #69, opened 2021-12-13
Some misc changes:

ashishmax31:
- Explicitly convert path object to string before calling .split()
  - Change yaml loader

Signed-off-by: ashish <asnaraya@redhat.com>

### DIFF
--- a/managedtenants/bundles/utils.py
+++ b/managedtenants/bundles/utils.py
@@ -122,7 +122,7 @@ def check_image_size(image_url_tag):
 def quay_org_from_path(quay_org_path, base_path="quay.io/"):
     # "quay.io/osd-addons/" -> osd-addons
     # "quay.io/osd-addons" -> osd-addons
-    split_res = quay_org_path.split(base_path)
+    split_res = str(quay_org_path).split(base_path)
     if split_res:
         return split_res[-1].strip("/")
     raise BundleUtilsError(
@@ -134,7 +134,7 @@ def quay_org_from_path(quay_org_path, base_path="quay.io/"):
 def load_yaml(path):
     try:
         with open(path, "r", encoding="utf8") as file_obj:
-            data = yaml.load(file_obj.read(), Loader=yaml.CSafeLoader)
+            data = yaml.load(file_obj.read(), Loader=yaml.SafeLoader)
             return data
     except yaml.YAMLError:
         return None


managedtenants/bundles/utils.py
```diff
@@ -122,7 +122,7 @@ def check_image_size(image_url_tag):
 def quay_org_from_path(quay_org_path, base_path="quay.io/"):
     # "quay.io/osd-addons/" -> osd-addons
     # "quay.io/osd-addons" -> osd-addons
-    split_res = quay_org_path.split(base_path)
+    split_res = str(quay_org_path).split(base_path)
```
yashvardhan-kukreja, line 125: anyplace where we noticed `quay_org_path` being non-string?
If there are such cases, isn't it better to `raise TypeError` instead, because a non-string passing through this function is non-intended behaviour and it might be better to get notified via errors, instead of non-string object getting converted into string (potentially falsely)?

wdyt

ashishmax31: Yes, `quay_org_path`would be path object. Thats what we pass to this library. (We pass`OSD_ADDONS_QUAY_ORG` 
from the mt-bundles repo as `quay_org_path `.)

managedtenants/bundles/utils.py
```diff
@@ -134,7 +134,7 @@ def quay_org_from_path(quay_org_path, base_path="quay.io/"):
 def load_yaml(path):
     try:
         with open(path, "r", encoding="utf8") as file_obj:
-            data = yaml.load(file_obj.read(), Loader=yaml.CSafeLoader)
+            data = yaml.load(file_obj.read(), Loader=yaml.SafeLoader)
```
ashishmax31, line 137: I was getting errors when testing out locally with `yaml.CSafeLoader`.

managedtenants/bundles/utils.py
yashvardhan-kukreja: Ohh gotcha! thanks :)

managedtenants/bundles/utils.py
yashvardhan-kukreja: Yep,  csafeloader ends up depending on libyaml bindings, so good catch :)

yashvardhan-kukreja (approved):
lgtm! Thanks for this :)

PR #71, opened 2021-12-13
Deprecate SingleNamespace installMode.

sugarraysam:


sugarraysam:
OCM supports single_namespace even though UI are not in sync: https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/pkg/api/addons.go#L43

PR #76, opened 2022-01-01
Add mtbundles schema for managed-tenants-bundles config.yaml

sugarraysam:
Add a schema for `config.yaml` file inside of managed-tenants-bundles.

Refactored a few constructs, and most importantly externalized the shared parts of `metadata`, `imageset` and `mtbundles` schemas as JSON pointers. (Could not get it to work with yaml...)


managedtenants/schemas/mtbundles.schema.yaml
```diff
@@ -0,0 +1,29 @@
...
+        $ref: "shared/addon_parameters.yaml"
+      addOnRequirements:
+        $ref: "shared/addon_requirements.yaml"
+      subOperators:
+        $ref: "shared/sub_operators.yaml"
+      subscriptionConfig:
+        $ref: "shared/subscription_config.json"
+required:
+  - environments
+  - ocm
```
ashishmax31, line 29: I don't think the ocm attribute is a required attribute

managedtenants/schemas/mtbundles.schema.yaml
```diff
@@ -0,0 +1,29 @@
...
+      enum:
+        - integration
+        - stage
+  ocm:
+    type: object
+    description: "Required OCM data for valid imagesets to be constructed by automated MRs to managed-tenants."
+    additionalProperties: false
+    properties:
+      addOnParameters:
+        $ref: "shared/addon_parameters.yaml"
```
ashishmax31, line 20: Did you mean `.json` here?

ashishmax31:
> Could not get it to work with yaml

Did some googling and it seems like we need an openAPI 3 parser to make it work for yaml.

managedtenants/schemas/mtbundles.schema.yaml
sugarraysam: nice catch

managedtenants/schemas/mtbundles.schema.yaml
sugarraysam: true, I suppose when we start using imagesets though it will become

sugarraysam:
> Did some googling and it seems like we need an openAPI 3 parser to make it work for yaml.

Interesting. Do you have a link? Did we want to use it? I am personally fine /w JSON. WDYT?

ashishmax31:
@sugarraysam I'm fine with json as well :)

ashishmax31 (approved):

PR #86, opened 2022-01-28
MTSRE-419 | fix bool misinterpretation

yashvardhan-kukreja:
the YAMLloader renders values in the jinja template without quotes. Now that perfectly works for generic strings but messes up when the value is something which can be interpreted as a non-string type like, true/false, any int like 2, 42, etc. 

But we want to deterministically ensure that under subscriptionConfig, the value, even if provided with "true", "false", "1", "2", etc. ends up explicitly getting interpreted as a string and the block chomping indicator `|-` in this PR accomplishes that

Example:

**Before this PR**

**_Regular succeeding situations_**

_Input_
```json
{"env": {"name": "healthy", "value": "yes"}}
```
_Intermediate Representation_
```yaml
env:
  name: healthy ### parsed by YAML as a string
  value: yes ### parsed by YAML as a string
```
_Dumped YAML Output_
```yaml
env:
  name: healthy
  value: yes
```

**_Failing situations_**

_Input_
```json
{"env": {"name": "healthy", "value": "true"}}
```
_Intermediate Representation_
```yaml
env:
  name: healthy ### parsed by YAML as a string
  value: true ### parsed by YAML as a bool despite being inputted with a stringified "true"
```
_Dumped YAML Output_
```yaml
env:
  name: healthy
  value: true ## bool
```

**After this PR**

_Input_
```json
{"env": {"name": "healthy", "value": "true"}}
```
_Intermediate Representation_
```yaml
env:
  name: healthy ### parsed by YAML as a string
  value: |-
    true ### parsed by YAML as a string
```
_Dumped YAML Output_
```yaml
env:
  name: healthy
  value: 'true' ## string
```

### DIFF
--- a/tests/testutils/addon_helpers.py
+++ b/tests/testutils/addon_helpers.py
@@ -69,10 +69,9 @@ def addon_with_indeximage():
 @pytest.fixture
 def addons_managed_by_addon_cr():
     def create_addon_managed_by_addon_cr(path):
-        addon = Addon(path, "stage")
-        addon.manager = AddonManager.ADDON_OPERATOR
-        addon.sss = Sss(addon=addon)
-        return addon
+        return Addon(
+            path, "stage", override_manager=AddonManager.ADDON_OPERATOR
+        )
 
     addon_paths = [addon_with_imageset_path(), addon_with_indeximage_path()]
     return list(map(create_addon_managed_by_addon_cr, addon_paths))
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,7 @@ repos:
       - id: trailing-whitespace
   # Black
   - repo: https://github.com/psf/black
-    rev: 21.12b0
+    rev: 22.1.0
     hooks:
       - id: black
         args: ["--config", ".linters/black", "--experimental-string-processing"]
--- a/managedtenants/core/addons_loader/addon.py
+++ b/managedtenants/core/addons_loader/addon.py
@@ -24,7 +24,7 @@ _PERMITTED_SUBSCRIPTION_CONFIGS = ["env"]
 
 
 class Addon:
-    def __init__(self, path, environment):
+    def __init__(self, path, environment, override_manager=None):
         self.path = path
         self.extra_resources_loader = None
         self.metadata = self.load_metadata(environment=environment)
@@ -57,6 +57,9 @@ class Addon:
             self.manager = AddonManager.UKNOWN
         else:
             self.manager = AddonManager.ADDON_OPERATOR
+
+        if override_manager is not None:
+            self.manager = override_manager
 
         self.sss = Sss(addon=self)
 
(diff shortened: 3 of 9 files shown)


tests/testutils/addon_helpers.py
```diff
@@ -69,10 +69,9 @@ def addon_with_indeximage():
 @pytest.fixture
 def addons_managed_by_addon_cr():
     def create_addon_managed_by_addon_cr(path):
-        addon = Addon(path, "stage")
-        addon.manager = AddonManager.ADDON_OPERATOR
-        addon.sss = Sss(addon=addon)
-        return addon
+        return Addon(
+            path, "stage", override_manager=AddonManager.ADDON_OPERATOR
```
sugarraysam, line 73: I am not a big fan of adding an function param simply to accomodate testing. Would suggest doing:

```python
addon = Addon(path, "stage")
addon.manager = AddonManager.ADDON_OPERATOR
return addon
```
```

**Review by sugarraysam — CHANGES_REQUESTED** [review]

File: tests/testutils/addon_helpers.py (hunk 1 above)
Comment (yashvardhan-kukreja) [reply tests/testutils/addon_helpers.py:73]: That would actually lead to unexpected results because
when the first line would execute, the `Addon(...)` constructor would initialise `addon.sss` assuming the manager to be `UNKNOWN` ([Ref](https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/core/addons_loader/addon.py#L56-L61
)) hence, the rendering the [wrong section](https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/data/selectorsyncset.yaml.j2#L9-L19) of the SSS Jinja template.

In fact, to counter this issue, till now (before this PR), the code was explicitly re-setting the `addon.sss` by doing the following 
```python
        addon.manager = AddonManager.ADDON_OPERATOR
        addon.sss = Sss(addon=addon)
```

Ref: https://github.com/mt-sre/managed-tenants-cli/blob/main/tests/testutils/addon_helpers.py#L73-L74

This was functionally correct but this was unnecessarily executing the SSS Walk execution twice (which happens everytime the `Sss()` constructor is called)

> Just to re-iterate: it will be called at `addon = Addon(...)` and `addon.sss = Sss(addon=addon)`

Hence, I felt the `Addon()` constructor should allow to provide/override the addon manager. 
And the `override_manager` argument has a default value (None) as well set in the [constructor's signature](https://github.com/mt-sre/managed-tenants-cli/blob/c2ce4d90421ddf88d0cf7ebb96fd02ac41fdc8f1/managedtenants/core/addons_loader/addon.py#L27) hence, doesn't require any change or even awareness about this change across the codebase.

sugarraysam: You are right, we've been rendering the Sss twice and my snippet was missing this info.

mayankshah1607 (approved):

sugarraysam (approved):

PR #90, opened 2022-02-08
MTSRE-433 | fix: versions to be sorted by Version Parser instead of them being lexicographically sorted

yashvardhan-kukreja:

### DIFF
--- a/managedtenants/bundles/bundle_builder.py
+++ b/managedtenants/bundles/bundle_builder.py
@@ -4,7 +4,6 @@ import subprocess
 from pathlib import Path
 
 import semver
-from packaging.version import parse as ParseVersion
 from sretoolbox.container import Image
 from sretoolbox.utils.logger import get_text_logger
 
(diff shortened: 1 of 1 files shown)


managedtenants/bundles/bundle_builder.py
```diff
@@ -4,6 +4,7 @@
 from pathlib import Path
 
 import semver
+from packaging.version import parse as ParseVersion
```
Ajpantuso, line 7: We already require `semver` so easier to use that. (See [here](https://github.com/mt-sre/managed-tenants-cli/blob/e29e76c215e5c63597f65b54d74220fc3a654287/setup.py#L61)).

You can also just use `max` with the `key` [keyword argument](https://python-semver.readthedocs.io/en/latest/usage/get-min-and-max-of-multiple-versions.html).

Ajpantuso (changes requested):
One comment regarding using `semver`, but the only other missing thing is a regression test. It's probably easiest to just make a helper called `max_version` which given a variadic sequence of strings will return the greatest as parsed by `semver`. Then just take the example from your description and make it the single testcase for a parameterized unit test.

managedtenants/bundles/bundle_builder.py
yashvardhan-kukreja: Ohh cool, thanks! Will do so :)

Ajpantuso (approved):
LGTM, we can add a unit test in a following PR.

PR #93, opened 2022-02-08
[MTSRE-401] Add annotations as labels to bundle images /w refactors

sugarraysam:
Here is a list of what is changed by this PR. It is a bit all over the place but tries to add more components and structure.

- add `--force-push` flag, that allow us to force-push images
- add `--debug` flag with extra logging
- always rebuild images (previous behavior if image exists in remote we don't rebuild, makes it difficult to troubleshoot)
- refactor some of ci.py as OO (work-in-progress)
- replace `subprocess.run("docker", ...)` /w docker-py library
- add authentication to dummy registry so we test the login workflow
- remove double validation, where we use operator_sdk to validate all local bundles, but also validate all remote images afterward

## Testing 

I ran the following commands successfully to validate this PR against prod bundles living in `managed-tenants-bundles`:

```py
$ (venv) $ managedtenants --addons-dir=../managed-tenants-bundles/addons --dry-run bundles
<built_everything_success>

$ (venv) $ managedtenants --addons-dir=../managed-tenants-bundles/addons --addon-name=gpu-operator --debug bundles --quay-org sblaisdo
<built_and_push_gpu_operator_to_my_repo>
```

## Issues

- We would like to use `mtcli list-bundles` to validate index images contain bundles, but it does not trust unsigned cert (can't enable in test), tried deploying full docker-compose + nginx /w TLS + IP SAN, but dont trust localhost: https://docs.docker.com/registry/recipes/nginx/
- Had to remove the `operator_sdk_validate(...)` right after building a bundle, because we already validate all bundle using their directories, and it fails to integrate with the login workflow


sugarraysam:
Seems we can't pull images from dockerhub. Trying to create a mirror in quay, but not seing any options to set repository state to Mirror ...

sugarraysam:
Found a quay container -> https://quay.io/repository/openshift/origin-docker-registry

managedtenants/bundles/docker_api.py
```diff
@@ -0,0 +1,174 @@
...
+
+        except QuayAPIError as e:
+            raise DockerError(
+                f"Failed to ensure quay repo {image.repository} got {e}."
+            )
+
+        except docker.errors.APIError as e:
+            raise DockerError(f"Failed to push {image.url_tag}, got {e}.")
+
+    def _login(self):
```
Ajpantuso, line 130: Did you try without logging in?

Basically login is only required if you don't have a config with credentials already populated.

Otherwise the docker-py client will just add the auth headers to each request from the cred store within the config file it loads.

It does expect the defined env key of `DOCKER_CONFIG` though:

https://github.com/docker/docker-py/blob/a48a5a9647761406d66e8271f19fab7fa0c5f582/docker/utils/config.py#L33-L38

The client doesn't expose a way to define the config location in the constructor so the file is either in an expected location or set by the ENV

Ajpantuso (approved):
LGTM

managedtenants/bundles/docker_api.py
sugarraysam: That worked like a charm :) thank you.

PR #95, opened 2022-02-11
Cleaning the house :)

sugarraysam:
Refactoring the `BundleBuilder` and adding other classes:

- `AddonBundles` - parses bundles related to an addon
- `Bundle` - represents a bundle

Added a bunch of tests as well.

No new functionalities, 100% refactors.


managedtenants/bundles/bundle.py
```diff
@@ -0,0 +1,101 @@
...
+
+import semver
+import yaml
+from sretoolbox.binaries import OperatorSDK
+
+from managedtenants.bundles.exceptions import BundleError
+
+
+class Bundle:
+    operator_sdk = OperatorSDK(version="1.4.2", download_path="/tmp")
```
Ajpantuso, line 12: Class variables are gross, so if you want to initialize this once and not have it as a dependency then declare as a private module constant outside the class: `_OPERATOR_SDK = OperatorSDK(version="1.4.2", download_path="/tmp")`. Doesn't make it immutable, but makes it much more difficult to accidentally modify.

managedtenants/bundles/bundle_builder.py
```diff
@@ -41,233 +31,34 @@ def __init__(
...
-        images = []
-        main_addon = self._is_main_addon(addon)
-        # The main addon lives inside the "main" directory, so
-        # we look at its parents name to get the addon name.
-        # (sblaisdo) MTSRE-225: prepend deps /w addon_id to avoid quay
-        # collisions
-        addon_id = addon.parent.name
-        addon_name = addon_id if main_addon else f"{addon_id}-{addon.name}"
+            self.log.info(f"Building {bundle}.")
+            _ = self.docker_api.build_bundle(bundle)
```
Ajpantuso, line 51: Why isn't there a `try..catch` here? If the intention is to just fail anyway then I don't see why `bundles` should be returned for pipelining.

managedtenants/bundles/bundle_builder.py
```diff
@@ -41,233 +31,34 @@ def __init__(
...
-    def get_all_operator_names(self):
-        all_bundles = self._get_all_bundles()
-        operator_names = []
-        for bundle in all_bundles:
-            operator_name = self.get_operator_name_from_bundle(bundle)
-            if operator_name:
-                operator_names.append(operator_name)
+        for bundle in bundles:
+            self.log.info(f"Pushing {bundle}.")
+            self.docker_api.push(bundle.image)
```
Ajpantuso, line 62: Same here regarding `try..catch`

managedtenants/bundles/ci.py
```diff
@@ -306,32 +286,33 @@ def __init__(self, args):
...
 
-            index_builder = self._index_builder(addon)
-            index_image = index_builder.build_push_index_image(
-                bundle_images=bundle_images,
-                hash_string=get_short_hash(),
+            bundles = self.bundle_builder.build_and_push_all(
+                bundles, get_short_hash()
+            )
+            index_image = self.index_builder.build_and_push(
+                bundles, get_short_hash()
```
Ajpantuso, line 297: This is the pipelining bit I mentioned above. It's useful if we skip failed bundles, but otherwise it just adds another check back to `BundleBuilder` to reason about which bundles are being included in the index.

Ajpantuso (changes requested):
Functionally looks okay, but have some interface questions.

managedtenants/bundles/addon_bundles.py
```diff
@@ -0,0 +1,81 @@
...
+        Returns the latest version amongst the main_bundles.
+        """
+        all_versions = [bundle.version for bundle in self.main_bundles]
+        return max(all_versions, key=semver.VersionInfo.parse)
+
+    def get_all_bundles(self):
+        """
+        Returns an addon's main_bundles and dependency_bundles.
+        """
+        return self.main_bundles + self.dependency_bundles
```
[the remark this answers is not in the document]
sugarraysam: fixed

managedtenants/bundles/bundle.py
sugarraysam: fixed

managedtenants/bundles/bundle_builder.py
sugarraysam: added try/catch, not returning bundles it is indeed not very useful.

managedtenants/bundles/bundle_builder.py
sugarraysam: fixed

managedtenants/bundles/ci.py
sugarraysam: removed pipelining

PR #97, opened 2022-02-11
New parameter type <resource_requirement>

jmolmo:
New parameter type <resource_requirement>

JIRA - SDA-5400

Introduced AddOnRequirements inside addon parameter options

Signed-off-by: Juan Miguel Olmo Martínez <jolmomar@redhat.com>

### DIFF
--- a/managedtenants/schemas/shared/addon_parameters.json
+++ b/managedtenants/schemas/shared/addon_parameters.json
@@ -32,7 +32,8 @@
           "number",
           "boolean",
           "cidr",
-          "resource"
+          "resource",
+          "resource_requirement"
         ]
       },
       "validation": {
@@ -73,6 +74,9 @@
             "value": {
               "type": "string",
               "format": "printable"
+            },
+            "addOnRequirements":{
+              "$ref": "addon_requirements.json"
             }
           }
         }
--- a/tests/testdata/addons/mock-operator-with-imagesets/metadata/stage/addon.yaml
+++ b/tests/testdata/addons/mock-operator-with-imagesets/metadata/stage/addon.yaml
@@ -29,5 +29,27 @@ addonImageSetVersion: 1.0.0
 # Default values for subscriptionConfig
 subscriptionConfig:
   env:
-  - name: DEFAULT
-    value: "TRUE"
+    - name: DEFAULT
+      value: "TRUE"
+# Test $ref references can be resolved in shared schemas
+addOnParameters:
+  - id: size
+    name: Managed StorageCluster size
+    description:
+      The size, in terabytes, of the Storage Cluster to be deployed. Currently
+      1 or 4 are supported.
+    value_type: resource_requirement
+    required: true
+    editable: true
+    enabled: true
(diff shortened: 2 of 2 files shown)


sugarraysam:
Would love to see some testing to make sure the resolver is able to resolve `$ref` in shared schemas.

Could you please modify https://github.com/mt-sre/managed-tenants-cli/blob/main/tests/testdata/addons/test-operator/metadata/stage/addon.yaml and add the `addOnParameters.options.addOnRequirements: <some_data_following_addon_req_schema>`

My concern is you have to modify `jsonschema`'s resolver to actually find those `$ref`. See here: https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/core/addons_loader/addon.py#L191

sugarraysam:
Indeed, the tooling version is still at `1.8.2` in managed-tenants.

So ok here's the plan:

1. Copy the new param to the testing addon: https://github.com/mt-sre/managed-tenants-cli/blob/main/tests/testdata/addons/test-operator/metadata/stage/addon.yaml. So it is part of the regular tests:
```yaml
  - name: 4 TiB
    value: '4'
    addOnRequirements:
      - id: managed_svc_machine_pool_req
        resource: machine_pool
        data:
          compute.cpu: 40
          compute.memory: 2147483648
        enabled: true
```

2. Release `py-mtcli`
3. Update release in `managed-tenants`: https://gitlab.cee.redhat.com/service/managed-tenants/-/blob/main/requirements-dev.txt#L5

jmolmo:
@sugarraysam : Thanks for the roadmap .. sorry but i am new in this area and most things are completelly new for me. Couls you clarify a little bit more?

> Copy the new param to the testing addon: tests/testdata/addons/mock-operator-with-bundles/metadata/stage/addon.yaml.
In the example you provide are not options which will use the $ref you have doubts .. is correct?
The modification will be more or less the same I have for the ocs converged addon, what will make the difference to have another mr with the same modification in other addon?

> Merge + release py-mtcli
I cannot merge any modification (i think no rights), and what is release py-mtcli?

Thanks and sorry to bother you with probably dumb questions

sugarraysam:
@jmolmo There is no such thing as a dumb question. 

First I will take care of merging + releasing. I just wanted to give you the roadmap before we can merge your other MR in `managed-tenants`.

We need testing to be integrated in this repository, so we catch any errors in development process. We want to avoid situations where we create a new release, to later find out it does not integrate in `managed-tenants`.

Please apply the same modification to your branch as found in the following commit. This is the only missing thing. This addon is loaded all over our testing suites, which will indirectly validate we can successfully parse it: https://github.com/sugarraysam/managed-tenants-cli/commit/e8041446be92888a441c17dae3f53dd7e6326887#diff-787e8e383a8091e7f88f202bfb79005da8af322a55477f35582862494daf593aR34-R51

jmolmo:
@sugarraysam : do i need to do something additional?

sugarraysam (approved):
lgtm

PR #101, opened 2022-02-16
[MTSRE-453] Allow only `latest` for addonImageSetVersion

mayankshah1607:

# py-mtcli

## Description

Modifed Addon loader to error out when `addonImageSetVersion` is set to anything apart from `"latest"`. See https://issues.redhat.com/browse/MTSRE-453
## Checklist

### DIFF
--- a/managedtenants/core/addons_loader/addon.py
+++ b/managedtenants/core/addons_loader/addon.py
@@ -24,13 +24,31 @@ _PERMITTED_SUBSCRIPTION_CONFIGS = ["env"]
 
 
 class Addon:
-    def __init__(self, path, environment, override_manager=None):
+    def __init__(
+        self,
+        path,
+        environment,
+        override_manager=None,
+        imageset_latest_only=False,
+    ):
         self.path = path
         self.extra_resources_loader = None
         self.metadata = self.load_metadata(environment=environment)
         self.imageset_version = self.metadata.get("addonImageSetVersion")
+        self.imageset_latest_only = imageset_latest_only
         # If imagebundles are provided
         if self.imageset_version is not None:
+
+            # Do not allow pin-pointing a version
+            # Temporary hot-fix for https://issues.redhat.com/browse/SDA-4918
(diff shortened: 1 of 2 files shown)


managedtenants/core/addons_loader/addon.py
```diff
@@ -111,6 +118,17 @@ def load_metadata(self, environment):
...
+        # Do not allow pin-pointing a version
+        # Temporary hot-fix for https://issues.redhat.com/browse/SDA-4918
+        # See: https://issues.redhat.com/browse/MTSRE-453
+        if self.imageset_latest_only and imageset_version != "latest":
+            raise AddonLoadError(
+                "Pointing to a specific version is not allowed. Please use"
+                " only latest. See https://issues.redhat.com/browse/MTSRE-453"
+                " for more info"
+            )
+
```
yashvardhan-kukreja, line 131: can we move this under the constructor itself i.e. under the line 
```py
...
if self.imageset_version is not None:
  if self.imageset_latest_only and imageset_version != "latest":
    raise AddonLoadError(
                "Pointing to a specific version is not allowed. Please use"
                " only latest. See https://issues.redhat.com/browse/MTSRE-453"
                " for more info"
            )
  self.imagesets_path = self.path / f"addonimagesets/{environment}"
  self.imageset = self.load_imageset(self.imageset_version)
  self.package = None
  self.bundles = None
  self.catalog_image = Image(self.imageset["indexImage"])
```

Semantically, it feels that leaving `load_imageset` to just "load imagesets" makes more sense and validation like this one should be a part of constructor (or maybe a separate method which gets called in the constructor) considering the fact that it relies upon setting up the Addon object itself more than the load_imageset step.

But this is my POV, not a hard requirement. So feel free to resolve this suggestion straight away if you think that  the original implementation makes more sense. I am happy with anything.

Rest looks good :)

mayankshah1607:
Thanks for the suggestion, @yashvardhan-kukreja ! I've made that change :)

yashvardhan-kukreja (approved):

PR #103, opened 2022-02-17
[MTSRE-415 + MTSRE-416] Enforce the config.yaml schema. Deduplication feature. Refactors.

sugarraysam:
## Scope

- MTSRE-415 - allow an `AddonBundles` to be used by multiple addons
- MTSRE-416 - enforce `mtbundles.schema.yaml` on the `config.yaml` file of `AddonBundles`
- create a single merge request and a single branch for all affected addons
- more OO refactors in ci.py
- Added `ImageSet` + `ImageSetCreator` classes
- extend `AddonBundles` to support `ImageSet` natively


managedtenants/bundles/addon_bundles.py
```diff
@@ -79,3 +136,74 @@ def get_all_bundles(self):
...
+    def get_all_imagesets(self, index_image):
+        """
+        Produce all imagesets for a given index_image.
+
+        :param index_image_str: representation of an index_image
+        """
+        res = []
+
+        version = self._get_latest_version()
+        ocm_config = self.config.get("ocm", {})
```
Ajpantuso, line 149:
```suggestion
        ocm_config = self._get_ocm_config()
```

Not sure, but it feels like that defaulting was supposed to be applied here.

Otherwise I don't see where this private method is used outside of tests.

managedtenants/bundles/bundle_builder.py
```diff
@@ -64,9 +65,17 @@ def _push_all(self, bundles):
...
             for bundle in bundles:
                 self.log.info(f"Pushing {bundle}.")
-                self.docker_api.push(bundle.image)
+                self.docker_api.push(
+                    bundle.image,
+                    ensure_repo=ensure_repo.get(
+                        bundle.bundle_repo_name(), True
+                    ),
+                )
+                ensure_repo[bundle.bundle_repo_name()] = False
```
Ajpantuso, line 78:
```suggestion
                repo = bundle.bundle_repo_name()
                self.docker_api.push(
                    bundle.image,
                    ensure_repo=repo not in seen,
                )
                seen.add(repo)
```

Using a set `seen` instead of a `dict` might be a little cleaner.

Ajpantuso (changes requested):
A couple of comments to get started, but looks okay for the most part.

managedtenants/bundles/addon_bundles.py
sugarraysam: good catch, experimented with various things and forget to switch it back

managedtenants/bundles/bundle_builder.py
sugarraysam: it does look cleaner. changed it !

managedtenants/bundles/addon_bundles.py
```diff
@@ -61,13 +75,56 @@ def _parse_operator_bundles(self, operator_dir):
...
+    def _parse_and_validate_config(self):
+        config_file = self.root_dir / "main" / "config.yaml"
+        try:
+            with open(config_file, "r", encoding="utf-8") as f:
+                config = yaml.safe_load(f)
+                self._validate_config(config)
+                return config
+
+        except FileNotFoundError:
+            # We are not enforcing the presence of config.yaml yet.
```
sugarraysam, line 99: This is not good. Forgot to change that back. We want to enforce the `config.yaml` workflow (pushed a fix).

sugarraysam:
@Ajpantuso PTAL

PR #106, opened 2022-02-21
make displayName allow OperatorHub style [DEPRECATED] strings

yashvardhan-kukreja:


managedtenants/data/selectorsyncset.yaml.j2
```diff
@@ -62,7 +62,8 @@ items:
         {{ maybe_annotations(ADDON.metadata.get('commonAnnotations')) | indent(8) }}
         {{ maybe_labels(ADDON.metadata.get('commonLabels')) | indent(8) }}
       spec:
-        displayName: {{ADDON.metadata['name']}}
+        displayName: |-
```
thetechnick, line 65:
```suggestion
        displayName: >|
```
Needs to be folded or we might have newlines in that string

yashvardhan-kukreja: Let's go ahead with `>-` then to chomp any possible newline chars in the end.

thetechnick: Ah sorry, copy-paste mistake. `>-` is right

thetechnick (approved):

PR #109, opened 2022-02-23
ocmCli changes

ashishmax31:
## Description
 - make `_api` method return raw response object on http requests
 - refactor `_addon_exists` method to return boolean values based on
   exception raised.

### DIFF
--- a/managedtenants/utils/ocm.py
+++ b/managedtenants/utils/ocm.py
@@ -15,10 +15,6 @@ class OCMAPIError(Exception):
         self.response = response
 
 
-class AddonCreationError(Exception):
-    pass
-
-
 def retry_hook(exception):
     if not isinstance(exception, OCMAPIError):
         return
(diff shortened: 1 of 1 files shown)


managedtenants/utils/ocm.py
```diff
@@ -107,16 +103,18 @@ def add_addon(self, metadata):
...
+        except OCMAPIError:
+            return False
 
     def add_addon_version(self, imageset, metadata):
         if self._addon_exists(metadata.get("id")):
             return self._add_addon_version(imageset, metadata)
         # Create the addon first
-        resp = self.add_addon(metadata)
-        if resp.status_code != 201:
-            raise AddonCreationError("Failed to succesfully create the addon.")
```
ashishmax31, line 119: We dont need a separate exception class here as `_post` method will raise `OCMAPIError` on failed requests (!= 201)

managedtenants/utils/ocm.py
```diff
@@ -107,16 +103,18 @@ def add_addon(self, metadata):
...
+        self.add_addon(metadata)
```
yashvardhan-kukreja, line 117: Let's handle some errors, for example, the way it's done here :

https://github.com/mt-sre/managed-tenants-cli/blob/415e8ab6a68e29e91f728bb9d4b2d757528f7729/managedtenants/utils/ocm.py#L159-L167

wdyt :)

ashishmax31: yep

ashishmax31: But there isn't any errors we can handle here, if the post call fails we just raise an exception. Its like how we do in other methods(where we just do ` return self._post("/api/clusters_mgmt/v1/addons", json=addon)`)

yashvardhan-kukreja: Ohh sorted then, if it's meant to fail, then, I am happy :) (never thought I'd say this sentence while reviewing a PR :P )

yashvardhan-kukreja (approved):
Thanks for your work :)
lgtm!

PR #111, opened 2022-02-23
[MTSRE-428] Implementing the single-bundle-per-operator pattern.

sugarraysam:
# py-mtcli

## Description

Implemented the single-bundle-per-operator pattern with local testing. Need to perform extended testing and refactoring in `managed-tenants-bundles`.

This will prevent us from accumulating bundle and speed up both the building side of things (`managed-tenants-bundles`) and the extraction + validation (`addon-metadata-operator`). It also facilitates migrating and upgrading our tooling as we don't have a zillion bundles to fix and maintain.

## Checklist

- [x] `$ managedtenants -addons-dir=../managed-tenants-bundles/addons --addon-name=<addon> bundles --quay-org <personal_org>` (successfuly built and push a single addon)


managedtenants/bundles/bundle.py
```diff
@@ -69,29 +89,44 @@ def _validate_operator_sdk(self):
...
-    def _parse_metadata_annotations(self):
-        annotations_file = self.path / "metadata" / "annotations.yaml"
-        with open(annotations_file, "r", encoding="utf-8") as f:
-            try:
-                data = yaml.load(f, Loader=yaml.CSafeLoader)
-
-            except FileNotFoundError as e:
-                raise BundleError(f"{self}: {annotations_file} not found: {e}")
+    def _validate_csv(self):
+        # 1. enforce the `olm.skipRange` annotation
```
sandhyadalavi, line 93: want to clarify. either skiprange or replaces should be present?, and this is enforcing skiprange to be present?

sugarraysam: It's a pattern where `replaces` can never be specified, and the `olm.skipRange` annotation has to be present. This is to make sure we define a proper upgrade graph from OLM's perspective. https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/

sugarraysam: This pattern is used by OCS currently (see their CSV's in managed-tenants-bundles).

managedtenants/bundles/csv.py
```diff
@@ -0,0 +1,58 @@
...
+        self.path = self._get_path()
+        self.data = self._parse_csv()
+
+    def _get_path(self):
+        """
+        Find the csv manifest path.
+        """
+        candidates = [
+            f"{self.operator_name}.csv.yaml",
+            f"{self.operator_name}.clusterserviceversion.yaml",
```
ashishmax31, line 19: I've noticed that some addon teams use `.yml` extensions as well. Do you think its a better idea to use regex here?
(I hate regex myself :D)

sugarraysam: Thanks for spotting this. Refactored to a double for loop.

sugarraysam: Re-refactored to only implement suffixes, there are too many name differences.

ashishmax31:
Thanks @sugarraysam

ashishmax31 (approved):

PR #112, opened 2022-03-03
OCM: Add credentials secret name for STS addons

vkareh:
This attribute is necessary to ensure that addon developers can use
a credentials secret name that matches what's expected by the operator.
OCM in turn will use this to create the correct credentials secret in
the customer cluster.


vkareh:
cc @ciaranRoche

vkareh:
Added docs here: https://gitlab.cee.redhat.com/service/managed-tenants/-/merge_requests/1843

PR #113, opened 2022-03-08
[MTSRE-472] Extending `operatorName` valid length

Ajpantuso:
# py-mtcli

## Description

- Updated metadata schema to permit `operatorname` longer than 32 characters.


sugarraysam:
This is a shaddy glitch with AMO, but can you re-run `make generate` to change the CRD definitions? I could not get it to work in `.pre-commit-config.yaml` there was a weird race condition.

sugarraysam:
Did you validate against current addons?

sugarraysam (approved):
LGTM

PR #126, opened 2022-04-04
[MTSRE-421] OPM 1.19.5

sugarraysam:
# py-mtcli

## Description

Bump OPM tooling to 1.19.5

- Abstracted the `LocalDockerRegistry` as I played with the idea of building/pushing images locally to workaround containerd cache not syncing with docker cache. It did not make it in the codebase, and is only used in testing, but leaving this construct there as we might want to revisit this option in the future.
- Bump sretoolbox for mtcli version parsing fix.
- Add `$ mtcli bundle validate <localDir>` to work locally and mimic `$ opm alpha bundle validate --tag <bundleImage>`. This allows a better workflow `validate - build -push`, as opposed to `build - push - validate` (opm only works with remote images).

## Checklist

- RHODS, RHOAMS and the gpu-operator bundles are broken, will tackle these in separate PRs, and eventually integrate a new version of `py-mtcli` in managed-tenants-bundles

- [x] `$ managedtenants -addons-dir=../managed-tenants-bundles/addons --addon-name=<addon> bundles --quay-org <personal_org>` (successfuly built and push a single addon)


managedtenants/bundles/docker_api.py
```diff
@@ -148,3 +142,9 @@ def _image_exists(self, image):
             # image.url_digest, calls digest which raises HTTPError
             # https://github.com/app-sre/sretoolbox/blob/master/sretoolbox/container/image.py#L119
             return False
+
+    def validate_image(self, tag):
+        image = self.client.images.get(tag)
```
sandhyadalavi, line 147: will there be a case, when images not presents?

sandhyadalavi: i dont think so and just want to confirm..

sugarraysam: In this case, we call this after building either a bundle image or an indexImage, so they should always be present.

sandhyadalavi (approved):
lgtm

PR #139, opened 2022-04-12
(fix) Lazy import bundles module.

sugarraysam:
# py-mtcli

## Description

Lazy import bundles module to avoid pulling binaries from the internet with sretoolbox unless they are required.


sandhyadalavi (approved):
lgtm

PR #140, opened 2022-04-13
[MTSRE-481] Fixed Indentation in the jinja template

rpriyanshu9:
# py-mtcli

## Description

Fixed indentation in the jinja template.

### DIFF
--- a/tests/sss/test_secrets.py
+++ b/tests/sss/test_secrets.py
@@ -19,7 +19,7 @@ def test_secrets(addon_with_secrets):
         for secret in sss_walker["template_deploy"]["objects"][0]["spec"][
             "resources"
         ]["Secret"]:
-            assert secret[1]["metadata"]["labels"] is not None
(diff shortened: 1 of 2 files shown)


tests/sss/test_secrets.py
```diff
@@ -16,5 +16,11 @@ def test_secrets(addon_with_secrets):
             )
             == no_of_secrets
         )
+        for secret in sss_walker["template_deploy"]["objects"][0]["spec"][
+            "resources"
+        ]["Secret"]:
+            assert secret[1]["metadata"]["labels"] is not None
```
yashvardhan-kukreja, line 22:
```suggestion
            assert secret[1].get("metadata", {}).get("labels") is not None
```
otherwise if metadata or labels is not found, Python would raise KeyError instead of just returning None.

yashvardhan-kukreja (approved):
/lgtm, 
thanks!

PR #141, opened 2022-04-13
LazyBin fix, pull binaries only when they are ran.

sugarraysam:
# py-mtcli

## Description

Delay pulling sretoolbox binaries to the moment when they are needed.

### DIFF
--- a/managedtenants/bundles/binary_deps.py
+++ b/managedtenants/bundles/binary_deps.py
@@ -1,5 +1,25 @@
 from sretoolbox.binaries import Mtcli, OperatorSDK, Opm
 
-OPM = Opm(version="1.19.5", download_path="/tmp")
-MTCLI = Mtcli(version="0.10.0", download_path="/tmp")
-OPERATOR_SDK = OperatorSDK(version="1.4.2", download_path="/tmp")
+
+class LazyBin:
+    """
+    Abstraction around sretoolbox to avoid pulling binaries from the internet
+    before they actually need to be ran.
+    """
+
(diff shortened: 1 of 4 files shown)


managedtenants/bundles/binary_deps.py
```diff
@@ -1,5 +1,25 @@
...
+    def run(self, cmd):
+        if self.instance is None:
+            self.instance = self.bin_class(self.version, self.download_path)
+
+        self.instance.run(*cmd)
+
+
+OPM = LazyBin(Opm, version="1.19.5", download_path="/tmp")
+MTCLI = LazyBin(Mtcli, version="0.10.0", download_path="/tmp")
+OPERATOR_SDK = LazyBin(OperatorSDK, version="1.4.2", download_path="/tmp")
```
sandhyadalavi, line 25: can we extract out these versions from here?

sandhyadalavi: other than that looks good

sugarraysam: I am sorry but I am not sure I understand. What do you mean?

sandhyadalavi: can we not hard-code these version here? extract out in a config file?

sugarraysam: I think extracting those versions to a config file, and giving a user the ability to change them at runtime is dangerous. The `managedtenants/bundles` package depends on these versions, and is only tested for these versions. I think we need to keep them hardcoded, and test any change with the whole PR->review->release process.

PR #147, opened 2022-04-22
Fix: Ensure SQL catalog contains bundles.

sugarraysam:
# py-mtcli

## Description

The codebase contained a bug where validating that a bundle image has Size > 0 is sufficient to ensure a bundle is properly built, but this is not sufficient for an index image.

This is because the index image will always have Size > 0 as it contains a whole alpine or busybox container depending on OPM's version used.

This fix proposes a deeper inspection by extracting the `/database/index.db` sqlite3 file from a container and ensuring the `properties` table contains bundles metadata.

## Checklist

### DIFF
--- a/managedtenants/bundles/docker_api.py
+++ b/managedtenants/bundles/docker_api.py
@@ -1,5 +1,7 @@
+import io
 import logging
 import os
+import tarfile
 
 import docker
 import docker.api.build
@@ -85,7 +87,7 @@ class DockerAPI:
             for log in log_generator:
                 self.log.debug(log)
 
-            self.validate_image(tag)
+            self.check_image_size_non_zero(tag)
             return out_image
 
         except docker.errors.BuildError as e:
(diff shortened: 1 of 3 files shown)


managedtenants/bundles/docker_api.py
```diff
@@ -143,8 +145,48 @@ def _image_exists(self, image):
...
+        try:
+            in_memory_tar = io.BytesIO()
+            container = self.client.containers.create(tag)
+            tar_bytes, _ = container.get_archive(path)
+
+            for tb in tar_bytes:
+                in_memory_tar.write(tb)
+
+            in_memory_tar.seek(0)
+            with tarfile.open(fileobj=in_memory_tar) as tf:
```
Ajpantuso, line 177:
```suggestion
            container = self.client.containers.create(tag)
            tar_bytes, _ = container.get_archive(path)

            with tarfile.open(fileobj=io.BytesIO(tar_bytes)) as tf:
```

I don't think you need the intermediary object. I would try initializing `BytesIO` with the constructor.

managedtenants/bundles/index_builder.py
```diff
@@ -88,3 +94,75 @@ def _push(self, index_image):
...
+
+    def _validate_sql_catalog(self, tag):
+        """
+        Validates an sql based index_image. Makes sure it contains bundles.
+        """
+        sql_catalog = None
+        try:
+            in_memory_db = self.docker_api.extract_file_from_container(
+                tag, "/database/index.db"
+            )
```
Ajpantuso, line 106: Can this be it's own `try` block for clarity?

managedtenants/bundles/index_builder.py
```diff
@@ -88,3 +94,75 @@ def _push(self, index_image):
...
+            err_msg = (
+                "failed to query bundles from '/database/index.db' inside of ",
+                f"{tag}, got {e}.",
+            )
+            self.log.error(err_msg)
+            raise IndexBuilderError(err_msg)
+
+        finally:
+            if sql_catalog:
+                sql_catalog.close()
```
Ajpantuso, line 134: Since this is a custom class I think adding `__enter__` and `__exit__` and using `SQLCatalog` as a context manager would keep things a little tidier.

Ajpantuso (commented):
A few comments, but looks good otherwise.

managedtenants/bundles/docker_api.py
sugarraysam: it did not like that

```py
Traceback (most recent call last):
  File "/home/sugar/opt/_mtrepos/managed-tenants-cli/.venv/bin/inspect_index_image.py", line 23, in <module>
    iibuilder._validate_sql_catalog(index_image)
  File "/home/sugar/opt/_mtrepos/managed-tenants-cli/managedtenants/bundles/index_builder.py", line 104, in _validate_sql_catalog
    in_memory_db = self.docker_api.extract_file_from_container(
  File "/home/sugar/opt/_mtrepos/managed-tenants-cli/managedtenants/bundles/docker_api.py", line 171, in extract_file_from_container
    with tarfile.open(fileobj=io.BytesIO(tar_bytes)) as tf:
TypeError: a bytes-like object is required, not 'generator'
```

managedtenants/bundles/index_builder.py
sugarraysam: fixed

managedtenants/bundles/index_builder.py
sugarraysam: fixed

sugarraysam:
Also realized the `container.remove(force=True)` was not executing and code was leaking containers. Not sure why, but moved it inside the `try...except`

Ajpantuso (approved):
LGTM

PR #149, opened 2022-04-26
[MTSRE-481] Added `destinationSecretName` field

rpriyanshu9:
# py-mtcli

## Description

- Added an optional `destinationSecretName` field to `addon.yaml` schema.
- Removed `vaultPath` regex.

### DIFF
--- a/managedtenants/data/selectorsyncset.yaml.j2
+++ b/managedtenants/data/selectorsyncset.yaml.j2
@@ -29,7 +29,7 @@ items:
       - apiVersion: v1
         kind: Secret
         metadata:
-          name: {{secret['name']}}
+          name: {{secret['destinationSecretName'] | default(secret['name'])}}
           namespace: {{ADDON.metadata['targetNamespace']}}
           {{ maybe_annotations(ADDON.metadata.get('commonAnnotations')) | indent(10) }}
           {{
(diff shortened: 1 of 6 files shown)


managedtenants/data/selectorsyncset.yaml.j2
```diff
@@ -29,7 +29,7 @@ items:
       - apiVersion: v1
         kind: Secret
         metadata:
-          name: {{secret['name']}}
+          name: {{secret['destinationSecretName'] | default(secret['name'])}}
```
yashvardhan-kukreja, line 32:
```suggestion
          name: {{secret.get('destinationSecretName') or secret['name']}}
```
shouldn't it be this? 

Considering the Pythonic nature of Jinja, I am afraid that `secret['destinationSecretName']` might raise KeyError instead of returning `None` just the way Python does?

Would love to be corrected in case you already tested this :)

rpriyanshu9: Yup, I already tested this. When `destinationSecretName` isn't defined, it just fallbacks to the default value, i.e. `secret['name']`. I can change it as per your suggestion too. What do you suggest?

yashvardhan-kukreja (approved):
lgtm!
Great job :)

PR #158, opened 2022-05-03
[MTSRE-528] serviceMonitor for AddonCR.

sugarraysam:
# py-mtcli

## Description

Feature parity with the addon-operator.

Depends on https://github.com/openshift/addon-operator/pull/186

### DIFF
--- a/tests/sss/test_federated_metrics.py
+++ b/tests/sss/test_federated_metrics.py
@@ -69,6 +69,8 @@ def test_namespace_and_servicemonitor_v2(data):
     env = data.draw(custom_strategies.environment())
     addon = data.draw(custom_strategies.addon(env))
     addon.metadata["monitoring"] = {
+        "portName": data.draw(custom_strategies.k8s_name()),
+        "namespace": data.draw(custom_strategies.namespace()),
         "matchNames": data.draw(
             hypothesis_strategies.lists(custom_strategies.k8s_name())
         ),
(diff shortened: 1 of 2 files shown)


tests/sss/test_federated_metrics.py
```diff
@@ -69,6 +69,8 @@ def test_namespace_and_servicemonitor_v2(data):
     env = data.draw(custom_strategies.environment())
     addon = data.draw(custom_strategies.addon(env))
     addon.metadata["monitoring"] = {
+        "portName": data.draw(custom_strategies.k8s_name()),
+        "namespace": data.draw(custom_strategies.namespace()),
```
sandhyadalavi, line 73: can we add matchlabels and matchnames here?

apahim (changes requested):
Can we remove the AddonCR from the SelectorSyncSet instead? It's not used anywhere, it's just causing confusion, and the feature parity work should happen here:

https://github.com/openshift/addon-operator/blob/main/internal/controllers/addon/phase_ensure_monitoring_federation.go

apahim:
I see you hve it here: https://github.com/openshift/addon-operator/pull/186/files

So, yeah, it all good, I believe. Unless OCM needs to pass some parameter through, as it's OCM who creates the Addon CR, not this SSS.

sugarraysam:
@apahim let me close this PR then and create a new one for the removal of the Addon CR.

PR #164, opened 2022-05-10
MTSRE-489 Bundles use digest format

rpriyanshu9:
# py-mtcli

## Description

Changed the tag format used by bundles to digest format.


managedtenants/bundles/index_builder.py
```diff
@@ -26,7 +26,11 @@ def __init__(
         )
 
     def build_and_push(
-        self, bundles, hash_string=get_short_hash(), skip_validation=False
+        self,
+        bundles,
+        hash_string=get_short_hash(),
+        skip_validation=False,
+        test_run=False,
```
yashvardhan-kukreja, line 33: Kinda not sold on the need of the extra arg of `test_run` to be honest where the `skip_validation` seems to be capable of accomplishing the same purpose :confused:
They both seem to duplicate each other, thereby unnecessarily complicating the API.

If semantics were your concern, then I think it would be better to just rename `skip_validation` to `dry_run` and use that instead but again, that also seems a bit far-fetched.

Or Am I missing something here and we do have a place where test_run might have a dedicated purpose like, `skip_validation=True` and `test_run=False` or vice versa?

Would love to know your thoughts on this and feel free to correct me wherever the need be :slightly_smiling_face:

Thanks!

managedtenants/bundles/index_builder.py
```diff
@@ -26,7 +26,11 @@ def __init__(
         )
 
     def build_and_push(
-        self, bundles, hash_string=get_short_hash(), skip_validation=False
+        self,
```
apahim, line 29: nitpick: refactoring the block and adding new params in the same commit makes review harder.

managedtenants/bundles/index_builder.py
```diff
@@ -35,11 +39,15 @@ def build_and_push(
...
+        index_image = self._build(
+            bundles, hash_string, skip_validation, test_run
+        )
         return self._push(index_image)
 
     # pylint: disable=unused-argument
-    def _build(self, bundles, hash_string, skip_validation=False):
+    def _build(
+        self, bundles, hash_string, skip_validation=False, test_run=False
+    ):
```
apahim, line 50: yep, pls group all refactorings in a separate commit, so reviews can go straight to what's important :)

managedtenants/bundles/index_builder.py
```diff
@@ -60,7 +68,14 @@ def _build(self, bundles, hash_string, skip_validation=False):
             "add",
             "--permissive",
             "--bundles",
-            ",".join([bundle.image.url_tag for bundle in bundles]),
+            ",".join(
+                [
+                    bundle.image.url_digest
+                    if not test_run
```
apahim, line 74: why do we not want to resolve the digest during testing?

tests/bundles/test_index_builder.py
```diff
@@ -24,7 +24,7 @@ def test_index_builder_build_and_push():
     # <registry_url>/v2/ and our quay.io registry doesn't support this workflow
     index_builder = IndexBuilder(docker_api=docker_api, debug=True)
     index_image = index_builder.build_and_push(
-        bundles, hash_string, skip_validation=True
+        bundles, hash_string, skip_validation=True, test_run=True
```
apahim, line 27: it seems like what you really need it to mock the registry.

managedtenants/bundles/index_builder.py
yashvardhan-kukreja: Thanks for pointing this out. I just noticed the `_validate_sql_catalog` part and `skip_validation` being dedicately used for that.
I guess the term "skip_validation" caused me to presume that it is more of a dry-run-oriented thing rather than specifically being dedicated to catalog validation.

Now I see the point of `test_run` although I would suggest a different for it. For example, instead of `test_run=False` maybe we can use `use_digest=True`.

```
              [
                    bundle.image.url_digest
                    if use_digest
                    else bundle.image.url_tag
                    for bundle in bundles
                ]
```
Because right now, the only use-case of test_run param is to signify to use url_tag or url_digest, instead of any diffenent test_run-oriented functional logic. (for example, skipping builds or modifying build procedure entirely).

In the future, if we wanna have a test_run-oriented logic in this method, then, we can either use an env var or a `dry_run` to signify that.

Let me know what do you think :)

yashvardhan-kukreja: Update on this: @rpriyanshu9 pointed out the actual reason behind the skip_validation param i.e. it is to skip the catalog-level validation of bundles.

yashvardhan-kukreja: Thanks for pointing this out. I just noticed the `_validate_sql_catalog` part and `skip_validation` being dedicately used for that.
I guess the term "skip_validation" caused me to presume that it is more of a dry-run-oriented thing rather than specifically being dedicated to catalog validation.

Now I see the point of `test_run` although I would suggest a different for it. For example, instead of `test_run=False` maybe we can use `use_digest=True`.

```
              [
                    bundle.image.url_digest
                    if use_digest
                    else bundle.image.url_tag
                    for bundle in bundles
                ]
```
Because right now, the only use-case of test_run param is to signify to use url_tag or url_digest, instead of any diffenent test_run-oriented functional logic. (for example, skipping builds or modifying build procedure entirely).

In the future, if we wanna have a test_run-oriented logic in this method, then, we can either use an env var or a `dry_run` to signify that.

Let me know what do you think :slightly_smiling_face:

apahim: My problem with test paths in the code is that you don't test what you run. instead, we should create mockups and execute the actual code. for instance, if a new version of the sretoolbox changes/breaks the `url_digest()`, we won't know until it's too late.

managedtenants/bundles/index_builder.py
rpriyanshu9: noted, thanks for the feedback :)

tests/bundles/test_index_builder.py
yashvardhan-kukreja: Update on this: We'll take a look at the LocalRegistry setup to see if we can modify it to match our use-case i.e. image's being accessible at `<local-test-registry-url>/v2`.

We have some ideas but we'll validate them tomorrow :) 

cc: @rpriyanshu9

apahim: Or it can use a local registry, as you mentioned. Personally, I'd go with the github-mirror example, for good-enough-with-less-dependencies, but any strategy is fine really.

yashvardhan-kukreja: Thanks a lot Amador for these references. Let's take a look at them to figure out the most convenient approach possible to meet our use-case :)

managedtenants/bundles/index_builder.py
rpriyanshu9: because the current testing registry does not provide any metadata i.e. the manifest that we require here: https://github.com/app-sre/sretoolbox/blob/86d49c7ac4d27aaf3dea4f179d2f02247285f120/sretoolbox/container/image.py#L123-L146

rpriyanshu9: update: On digging into the tests we have in place, I found that the bundles weren't getting pushed in the test registry. Logs:
```
INFO     managedtenants-bundle-builder:bundle_builder.py:68 Pushing Bundle(addon_name=reference-addon, operator_name=reference-addon, version=0.1.6, path=/home/prraj/Desktop/redhat-projects/managed-tenants-cli/tests/testdata/addons/reference-addon/main/0.1.6, image=127.0.0.1:9999/reference-addon-bundle:0.1.6-0aa5ea5d8c3c4d1989d3d91ae3048e41).
DEBUG    managedtenants-docker:docker_api.py:120 {'status': 'The push refers to repository [127.0.0.1:9999/reference-addon-addon-operator-bundle]'}
DEBUG    managedtenants-docker:docker_api.py:120 {'errorDetail': {'message': 'Get "http://127.0.0.1:9999/v2/": read tcp 127.0.0.1:33606->127.0.0.1:9999: read: connection reset by peer'}, 'error': 'Get "http://127.0.0.1:9999/v2/": read tcp 127.0.0.1:33606->127.0.0.1:9999: read: connection reset by peer'}
```
Working on fixing this too.

rpriyanshu9: This is fixed as we use quay.io/mtsre/registry as the `LocalDockerRegistry` now.

tests/bundles/conftest.py
```diff
@@ -7,7 +7,7 @@
 from managedtenants.data.paths import BUNDLES_DIR
 
 LOCAL_REGISTRY_NAME = "mt-bundles-test-registry"
-LOCAL_REGISTRY_PORT = 9999
+LOCAL_REGISTRY_PORT = 5000
```
yashvardhan-kukreja, line 10:
```suggestion
LOCAL_REGISTRY_PORT = 443
```

tests/bundles/conftest.py
```diff
@@ -7,7 +7,7 @@
...
 REGISTRY_URL = f"127.0.0.1:{LOCAL_REGISTRY_PORT}"
```
yashvardhan-kukreja, line 11:
```suggestion
REGISTRY_URL = f"127.0.0.1:443"
```

yashvardhan-kukreja (approved):
lgtm, thanks a lot for putting this together

PR #165, opened 2022-05-11
[MTSRE-534] Create CatlogSource in the targetNamespace

Ankit152:
Signed-off-by: Ankit Kurmi <akurmi@redhat.com>

# py-mtcli

## Description

- Currently the SSS creates the CatalogSource in the openshift-marketplace namespace. We should use the targetNamespace instead:
https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/data/selectorsyncset.yaml.j2#L117

- addon-operator already does the right thing:
https://github.com/openshift/addon-operator/blob/main/internal/controllers/addon/phase_ensure_catalogsource.go#L33-L49

Added fix for the jinja template.

### DIFF
--- a/managedtenants/data/selectorsyncset.yaml.j2
+++ b/managedtenants/data/selectorsyncset.yaml.j2
@@ -128,7 +128,7 @@ items:
         channel: {{ADDON.metadata['defaultChannel']}}
         name: {{ADDON.metadata['id']}}
         source: addon-{{ADDON.metadata['id']}}-catalog
-        sourceNamespace: openshift-marketplace
+        sourceNamespace: {{ADDON.metadata['targetNamespace']}}
         {% if ADDON.get_subscription_config() %}
         config:
           env:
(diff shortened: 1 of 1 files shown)


managedtenants/data/selectorsyncset.yaml.j2
```diff
@@ -66,7 +66,7 @@ items:
       kind: CatalogSource
       metadata:
         name: addon-{{ADDON.metadata['id']}}-catalog
-        namespace: openshift-marketplace
+        namespace: {{ADDON.metadata["targetNamespace"]}}
```
apahim, line 69: we also have to change the `sourceNamespace` reference to this CatalogSource in the Subscription

apahim:
We need proper verification here, test this change with one addon first, just to make sure.

apahim:
This can be merged. PLease make sure to promote it today. @ashishmax31 @yashvardhan-kukreja

yashvardhan-kukreja (approved):
thanks for this @Ankit152 ! 

lgtm

Ankit152:
Thanks @yashvardhan-kukreja @apahim and @ashishmax31 🙂

PR #175, opened 2022-05-19
[MTSRE-538] feat: adding new command entrypoint for imageset validation

Ajpantuso:
# py-mtcli

## Description

- Adds a new command for validating imagesets to be used by CPaaS in a container.


sugarraysam (approved):
LGTM

PR #179, opened 2022-06-01
MTSRE-559 | Validates keys are unique in addon.yaml

rpriyanshu9:
Signed-off-by: Priyanshu Raj <prraj@redhat.com>

# py-mtcli

https://issues.redhat.com/browse/MTSRE-559


Ajpantuso (changes requested):
Other then the one comment I think a test case for this would be good as well.

Ajpantuso (approved):
LGTM

PR #184, opened 2022-06-08
Update ocm client

ashishmax31:
# py-mtcli

## Description
Updates ocm client to include the following attributes:
- additionalCatalogSources
- pullSecretName
- subscriptionConfig

Attributes defined in the imageset file takes precedence over the attributes defined in the metadata file.

### DIFF
--- a/managedtenants/utils/ocm.py
+++ b/managedtenants/utils/ocm.py
@@ -273,33 +273,40 @@ class OcmCli:
         # They will get overwritten with the values from the imageset if they're
         # present in the imageset as well.
         if metadata.get("pullSecretName"):
-            addon[self.IMAGESET_KEYS["pullSecretName"]] = metadata.get("pullSecretName")
+            mapped_key = self.IMAGESET_KEYS["pullSecretName"]
+            addon[mapped_key] = metadata.get("pullSecretName")
 
         if metadata.get("additionalCatalogSources"):
-            addon[self.IMAGESET_KEYS["additionalCatalogSources"]] = self._make_additional_ctlg_src_list(
+            mapped_key = self.IMAGESET_KEYS["additionalCatalogSources"]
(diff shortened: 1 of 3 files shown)


managedtenants/utils/ocm.py
```diff
@@ -277,6 +316,15 @@ def _addon_from_imageset(self, imageset, metadata):
                 addon[self.IMAGESET_KEYS[key]] = val
         return addon
 
+    @staticmethod
+    def _make_additional_ctlg_src_list(ctlg_srcs):
```
garrettthomaskth, line 320: Could `_make_additional_ctlg_src_list` and`_make_env_vars_list` be combined into a single function?

Ajpantuso: It would make a nice one liner too:
```python
def index_dicts(dicts):
    return [dict(d, id=str(id)) for id, d in enumerate(dicts)]
```

managedtenants/utils/ocm.py
```diff
@@ -265,8 +269,43 @@ def _addon_from_imageset(self, imageset, metadata):
...
+            addon[self.IMAGESET_KEYS["subscriptionConfig"]] = {}
+            addon[self.IMAGESET_KEYS["subscriptionConfig"]][
+                "add_on_environment_variables"
+            ] = self._make_env_vars_list(
+                metadata.get("subscriptionConfig").get("env")
+            )
+
         for key, val in imageset.items():
             if key in self.IMAGESET_KEYS:
+                if key == "additionalCatalogSources":
```
garrettthomaskth, line 297: This is a very small thing but it might be a little bit easier to read if we assigned `self.IMAGESET_KEYS[key]` to a variable, ex. `mapped_key = self.IMAGESET_KEYS[key]`

PR #185, opened 2022-06-08
Addons version and fields to secrets schema

apahim:
We need to add each field individually. Also, we have to validate them agains app-interface.


rpriyanshu9:
LGTM

yashvardhan-kukreja (approved):

PR #194, opened 2022-07-01
[MTSRE-594] validation for additionalCatalogSource Images

rpriyanshu9:
Signed-off-by: Priyanshu Raj <prraj@redhat.com>

# py-mtcli

## Description

Added validation check for additionalCatalogSource images to be present in digest format.
https://issues.redhat.com/browse/MTSRE-594

### DIFF
--- a/managedtenants/schemas/metadata.schema.yaml
+++ b/managedtenants/schemas/metadata.schema.yaml
@@ -143,7 +143,8 @@ properties:
           description: "Name of the additional catalog source."
         image:
           type: string
-          description: "Url of the additional catalog source image."
+          pattern: ([\w\-]+[.][\w\-.]+([:][0-9]+)?\/)([\w\-]+\/)?[\w\-.]+(@sha256:[0-9a-f]{64})$
+          description: "Url of the additional catalog source image in digest format."
       required:
         - name
         - image
--- a/managedtenants/schemas/metadata.schema.yaml
+++ b/managedtenants/schemas/metadata.schema.yaml
@@ -143,7 +143,7 @@ properties:
           description: "Name of the additional catalog source."
         image:
           type: string
-          pattern: ([\w\-]+[.][\w\-.]+([:][0-9]+)?\/)([\w\-]+\/)?[a-zA-Z0-9][\w\-.]+[a-zA-Z0-9](@sha256:[0-9a-f]{64})$
+          pattern: .*@sha256:[0-9a-f]{64}$
           description: "Url of the additional catalog source image in digest format."
       required:
         - name
--- a/managedtenants/schemas/metadata.schema.yaml
+++ b/managedtenants/schemas/metadata.schema.yaml
@@ -143,7 +143,7 @@ properties:
           description: "Name of the additional catalog source."
         image:
           type: string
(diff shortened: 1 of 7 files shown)


apahim:
Can't we just add a `pattern` here?

https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/schemas/metadata.schema.yaml#L144-L146

managedtenants/schemas/metadata.schema.yaml
```diff
@@ -143,7 +143,8 @@ properties:
           description: "Name of the additional catalog source."
         image:
           type: string
-          description: "Url of the additional catalog source image."
+          pattern: ([\w\-]+[.][\w\-.]+([:][0-9]+)?\/)([\w\-]+\/)?[\w\-.]+(@sha256:[0-9a-f]{64})$
```
rpriyanshu9, line 146: @apahim done! Please take a look at the pattern once.

rpriyanshu9: Or maybe we can go with just `(@sha256:[0-9a-f]{64})$`, what do you suggest? :)

managedtenants/schemas/metadata.schema.yaml
```diff
@@ -143,7 +143,8 @@ properties:
           description: "Name of the additional catalog source."
         image:
           type: string
-          description: "Url of the additional catalog source image."
+          pattern: ([\w\-]+[.][\w\-.]+([:][0-9]+)?\/)([\w\-]+\/)?[a-zA-Z0-9][\w\-.]+[a-zA-Z0-9](@sha256:[0-9a-f]{64})$
```
rpriyanshu9, line 146: done @apahim! PTAL at the pattern. I took some help from here: https://github.com/app-sre/sretoolbox/blob/97cc4db175086584cf2e9856bb58dfbba1e8aa57/sretoolbox/container/image.py#L282-L300

apahim: Hi @rpriyanshu9, nice that you knew about that regex! It's super complex, because it's meant to be a parser, to isolale each component of the URL.

I believe we need something very simplistic here, just to make sure it's a digest, but also easy to maintain.

I think it would be enough to have something like ".*@sha256:.*". What do you think? I know, it must have tekaen you quite some time to come up with that regex, but that's exactly my point :D

rpriyanshu9: Hey, thanks for your feedback. I agree with your point. What do you think about this, `.*@sha256:[0-9a-f]{64}$`?

rpriyanshu9: - https://github.com/mt-sre/managed-tenants-cli/blob/81f5ab059ca9e692121f072f6bf0e4767e823800/tests/testdata/addons/mock-operator-with-secrets/metadata/stage/addon.yaml#L47
- https://github.com/mt-sre/managed-tenants-cli/blob/81f5ab059ca9e692121f072f6bf0e4767e823800/tests/testdata/addons/test-operator/metadata/integration/addon.yaml#L22

apahim: Fair enough :+1:

apahim (changes requested):
There's a bigger ongoing discussion, to understand if this will really be merged, I let you know when we are settled there. For now, new just keep this as a draft please.

PR #206, opened 2022-07-11
fix: fail early when addon paramter has disallowed property combinations

Ajpantuso:
# py-mtcli

## Description

Adds the following validations for addon parameters:

- If `editable_direction` is defined, then `options` must be defined and non-empty
- For every `option` _o_, if _o.rank_ is defined, then _o.rank_ must be non-negative


Ajpantuso:
> LGTM

Can you leave an approving review as well? (It's required for merge)

apahim (approved):
lgtm

PR #228, opened 2022-07-25
Update ocm client

ashishmax31:
Changes:
- Allow default values to be picked up for addon requirements and suboperators when posting to OCM's addon version endpoint
- Set default empty values for addon secrets and envs when posting to OCM's addons endpoint, so that when existing secrets/envs are deleted from the addon.yaml files they're deleted from OCM as well.
- When creating new imagesets dont set default values for ocm related attrs


managedtenants/utils/ocm.py
```diff
@@ -304,6 +304,14 @@ def _addon_from_imageset(self, imageset, metadata):
                 metadata.get("addOnParameters")
             )
 
+        if metadata.get("addOnRequirements"):
+            mapped_key = self.IMAGESET_KEYS["addOnRequirements"]
+            addon[mapped_key] = metadata.get("addOnRequirements")
+
+        if metadata.get("subOperators"):
+            mapped_key = self.IMAGESET_KEYS["addOnRequirements"]
```
garrettthomaskth, line 312:
```suggestion
            mapped_key = self.IMAGESET_KEYS["subOperators"]
```

ashishmax31: Thanks

managedtenants/bundles/addon_bundles.py
```diff
@@ -185,17 +185,20 @@ def _get_ocm_config(self):
...
+            return ocm_config.get(k)
 
         return {
-            k: ocm_config_getter(k, default)
-            for k, default in [
-                ("addOnParameters", []),
-                ("addOnRequirements", []),
-                ("subOperators", []),
-                ("config", {}),
+            k: ocm_config_getter(k)
```
garrettthomaskth, line 192: This is nit-picky, but I think you could use an assignment expression here! See point 4 of the first answer https://stackoverflow.com/questions/10291997/how-can-i-do-assignments-in-a-list-comprehension

garrettthomaskth: I think it would look like 
`{k: value for k in [...] if (value := ocm_config.get(k))`

managedtenants/utils/ocm.py
```diff
@@ -322,15 +330,17 @@ def _addon_from_imageset(self, imageset, metadata):
...
     def set_addon_config(self, addon, config_obj, mapped_key):
-        if config_obj:
+        if config_obj is not None:
             if not addon.get(mapped_key):
                 addon[mapped_key] = {}
 
-            if config_obj.get("env"):
-                env_var_list = self.index_dicts(config_obj["env"])
+            # Can be an empty list hence the not none check
+            if config_obj.get("env") is not None:
```
garrettthomaskth, line 338: Also nit-picky, but what about `if "env" in config_obj:`?

managedtenants/bundles/addon_bundles.py
Ajpantuso: It was in our style guide to avoid assignment expressions since they force Python `>=3.8` (which doesn't apply anymore), but IMO this would be more readable if it was less compact.

Something like:

```python
...
raw_ocm_config = self.config.get('ocm', {})

_OCM_CONFIG_FIELDS = (
    "addOnParameters",
    "addOnRequirements",
    "subOperators",
    "config",
    "pullSecretName",
    "additionalCatalogSources",
)

return _remove_empty_values(
    _select_fields(_OCM_CONFIG_FIELDS, raw_ocm_config)
)

...

def _select_fields(fields, d):
    return {k: v for k, v in d.items() if k in fields}
    
    
def _remove_empty_values(d):
    return {k: v for k, v in d.items() if v}
```

ashishmax31: Thanks guys

ashishmax31: we dont want to remove explicitly set empty values

ashishmax31:
@Ajpantuso @garrettthomaskth please have a look now

ashishmax31:
@Ajpantuso @garrettthomaskth Please have look now

Ajpantuso (approved):
LGTM

PR #233, opened 2022-08-02
bundle builder: accept imageset enabled addons as a command line arg.

ashishmax31:
Accept a new command line parameter which lists the addons for which imageset has to be created when building index images.
Signed-off-by: ashish <asnaraya@redhat.com>


sandhyadalavi (approved):
LGTM

PR #245, opened 2022-08-23
Feat/ocm shortcircuit migrate

ashishmax31:
# py-mtcli

## Description
Adds a new method called `shortcircuit_migrate` to the ocm client class which creates the migration object, enables it and then whitelists it, all wrapped in timeouts and max tries.

### DIFF
--- a/managedtenants/utils/general_utils.py
+++ b/managedtenants/utils/general_utils.py
@@ -1,4 +1,7 @@
+import signal
 import subprocess
+from contextlib import contextmanager
+from time import sleep
 
 import semver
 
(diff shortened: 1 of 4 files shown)


managedtenants/utils/general_utils.py
```diff
@@ -23,3 +26,34 @@ def run(cmd, logger=None):
...
+
+    signal.signal(signal.SIGALRM, timeout_handler)
+    signal.alarm(duration)
+    try:
+        yield
+    finally:
+        signal.alarm(0)
+
+
+def try_with_timeout_until(
```
apahim, line 44: looks a lot like the sretoolbox.utils.retry https://github.com/app-sre/sretoolbox/blob/master/sretoolbox/utils/retry.py

ashishmax31: @apahim I'm not really sure the retry API provided in the sre toolbox is right here. The retry provided in the sre toolbox works on exceptions, whereas we want retry on predicates.

apahim: fair enough

managedtenants/utils/ocm.py
```diff
@@ -162,6 +165,36 @@ def disable_addon(self, addon_id):
             f"/api/clusters_mgmt/v1/addons/{addon_id}", json={"enabled": False}
         )
 
+    def shortcircuit_migrate(self, addon_id):
+        def addon_migration_complete():
+            migration = self.get_addon_migration(addon_id).json()
+            return migration.get("state") == "completed"
```
yashvardhan-kukreja, line 171:
```suggestion
        addon_migration_complete = self.get_addon_migration(addon_id).json().get("state") == "completed"
```
wouldn't this be cleaner.
I don't understand how this function specifically helps :)

yashvardhan-kukreja, line 171:
```suggestion
        addon_migration_complete = lambda _: self.get_addon_migration(addon_id).json().get("state") == "completed"
```

imo we can use the benefits of lambda function here :)

ashishmax31: I thought of using it as well. But the linter is not happy about the line length and as far as i know there is no way to split a lambda function into multiple lines in python. :/

yashvardhan-kukreja: ahh gotcha, we could skip the linter check but ehh too much hassle.

yashvardhan-kukreja (approved):

PR #258, opened 2022-09-16
[MTSRE-692] ocm client send namespace with labels

Ankit152:
Signed-off-by: Ankit Kurmi <akurmi@redhat.com>

# py-mtcli

## Description

ocm client to send in namespaces(along with labels) along with the addon payload


ashishmax31:
Move the logic to  ` _addon_from_metadata` (Maybe define a new method which returns the namespace list and use it)

managedtenants/utils/ocm.py
```diff
@@ -427,6 +436,10 @@ def _addon_from_metadata(self, metadata):
         metadata["config"] = metadata.get("config", {})
         metadata["config"]["env"] = metadata["config"].get("env", [])
         metadata["config"]["secrets"] = metadata["config"].get("secrets", [])
+        # ocm client to send in namespaces(along with labels)
+        namespaceLabels = metadata.get("namespaceLabels",{})
```
ashishmax31, line 440: Dont send `{}` if labels are missing, omit them altogether

managedtenants/utils/ocm.py
```diff
@@ -427,6 +436,10 @@ def _addon_from_metadata(self, metadata):
...
+        namespaces = metadata.get("namespaces",[])
+        metadata["namespaces"] = self.get_namespace_labels_list(namespaces,namespaceLabels)
 
         for key, val in metadata.items():
             if key in self.ADDON_KEYS:
```
ashishmax31, line 445: You need to add the `namespaces` key to `ADDON_KEYS`

ashishmax31 (changes requested):
Also add a test to https://github.com/mt-sre/managed-tenants-cli/blob/1629acc34da2cdf06cec915af957feea789c3d70/tests/utils/test_ocm.py

managedtenants/utils/ocm.py
```diff
@@ -50,6 +50,8 @@ class OcmCli:
         "subOperators": "sub_operators",
         "managedService": "managed_service",
         "config": "config",
+        "namespaces": "namespaces",
+        "namespaceLabels": "namespaceLabels",
```
ashishmax31, line 54: `namespaceLabels` is not required here as we don't send that to ocm. Otherwise looks good

Ankit152: cool. removing it. 🙂

ashishmax31 (approved):
lgtm

PR #277, opened 2022-10-27
[MTSRE-712] fix: index addon parameters starting from 1

Ajpantuso:
### Summary

Clusters-service ignores `order` updates to addon parameters with `order=0` since `0` is a useful default in go. This starts all parameter lists at index `1` so that updates by tenants are actually reflected in OCM.


apahim (approved):
lgtm

apahim:
great catch!

PR #279, opened 2022-10-28
Update Tooling to point to new API MTSRE-601

chamalabey:
# py-mtcli

## Description

Update Tooling to point to new API

- modified create/update _api/addons_mgmt/v1/addons/_ 
- modified create/update _api/addons_mgmt/v1/addons/{addon_name}/versions/_

Further details in [MTSRE-601](https://issues.redhat.com/browse/MTSRE-601)

## Checklist


chamalabey:
I have two concerns/questions 
- How do we handle the exceptions when calling two API endpoints for the same operation? 
- What if one of the calls for endpoints fails, the data in manage-service and addon-service will be out of sync.  How are we going to address those instances?

yashvardhan-kukreja:
> What if one of the calls for endpoints fails, the data in manage-service and addon-service will be out of sync. How are we going to address those instances?

Exactly what I was about raise @chamalabey , I believe that the lack of atomicity is a serious concern. I believe that implementing rollback workarounds would just incur a lot of engineering load that too with a not-so-elegant solution.

Let's maybe try to reach out to service-development folks (working actively on cluster/addons-service) and ask them if they can introduce such atomic guarantees in some way

managedtenants/utils/ocm.py
```diff
@@ -111,7 +111,11 @@ def list_sku_rules(self):
 
     def add_addon(self, metadata):
         addon = self._addon_from_metadata(metadata)
-        return self._post("/api/clusters_mgmt/v1/addons", json=addon)
+        # Update Tooling to point to new API MTSRE-601
+        self._post(
+            "/api/addons_mgmt/v1/addons", json=addon)
+        return self._post(
+            "/api/clusters_mgmt/v1/addons", json=addon)
```
yashvardhan-kukreja, line 118: I see that for every change you have followed the order:
1. addons_mgmt endpoint
2. clusters_mgmt endpoint

Now, as we addressed above, this can compromise atomicity say, when the Step 1 succeeds and Step 2 fails i.e. the addons_mgmt endpoint will be updated with new data whereas cluster_mgmt one will be left with stale.

In theory, we can swap that so that in the worst case, cluster_mgmt ends up with new data and addons_mgmt ends up with stale.

I believe we should have a chat with service-dev folks to see which of the above two situations would be more reliable.

If the service-dev folks have implemented some sort of reconciliation between addons_mgmt and cluster_mgmt (say, addons_mgmt periodically syncs itself to clusters_mgmt), then we can go with
Step 1. clusters_mgmt
Step 2. addons_mgmt

yashvardhan-kukreja:
> How do we handle the exceptions when calling two API endpoints for the same operation?

I believe raising an exception whenever even one of them fails should be suitable.

managedtenants/utils/ocm.py
chamalabey: Thank you Yash.
How do I contact service-dev folks to ask above questions?

chamalabey:
> > How do we handle the exceptions when calling two API endpoints for the same operation?
> 
> I believe raising an exception whenever even one of them fails should be suitable. I can see a situation where we might be happy with atleast one of the API endpoints succeeding but that would be only acceptable when the upstream OCM API has a reconciliation mechanism setup to synchronize clusters_mgmt with addons_mgmt periodically.
> 
> So, I say let's go raising with an exception even if one of the API endpoints fail.

Does that mean for now we should revert the change for the endpoint which succeeded? (e. g. created addon can be removed)
But addon version update rollback will be much more complicated. 
Just raising and exception will not address the data discrepancy, which is worrying.

ashishmax31:
@yashvardhan-kukreja @chamalabey 
> If the service-dev folks have implemented some sort of reconciliation between addons_mgmt and cluster_mgmt (say, addons_mgmt periodically syncs itself to clusters_mgmt

There is no such mechanism. Hitting these two endpoints is only going to be a interim solution, once we have AS ready and tested, we need to only hit the AS endpoint.
As for the data discrepency that can arise at the moment, sda already has a ticket that tracks that https://issues.redhat.com/browse/SDA-7055

managedtenants/utils/ocm.py
yashvardhan-kukreja: we can reach out to #service-development

chamalabey:
> @yashvardhan-kukreja @chamalabey
> 
> > If the service-dev folks have implemented some sort of reconciliation between addons_mgmt and cluster_mgmt (say, addons_mgmt periodically syncs itself to clusters_mgmt
> 
> There is no such mechanism. Hitting these two endpoints is only going to be a interim solution, once we have AS ready and tested, we need to only hit the AS endpoint. As for the data discrepency that can arise at the moment, sda already has a ticket that tracks that https://issues.redhat.com/browse/SDA-7055

Thank you @ashishmax31. My understanding is that SDA-7055 does not address the data discrepancy and it only perform the comparison of data between CS and AS

pepedocs:
> 

This is my recommendation as well (see jira question). I do not see the value (yet) on calling 2 APIs and as clients of these APIs we should only care about calling the right API and handling its result whatever that is. Normally in my experience API deprecation should not require 2 APIs to be called, it's just a matter of switching from an old API to a new API and don't care about what happens in the background - clients shouldn't need to know this. So I don't get the value of calling 2 APIs again. If we are going to delay this then we have to make sure we communicate this to the other stakeholders. I've started a comment in Jira.

yashvardhan-kukreja:
@chamalabey instead of injecting AS-related API calls into the existing methods calling CS-APIs, can you define two new methods: `addon_service_upsert_addon()` and `addon_service_upsert_addon_version()` where each of them insert/updates (i.e. upsert) addon and addon-version against AS respectively. Exactly like how it's defined [here](https://github.com/mt-sre/managed-tenants-cli/blob/947ffa3a8b31544db66018d281a26b06d9579d36/managedtenants/utils/ocm.py#L312-L330
) for CS-related API calls which happen today?

Thank you and feel free to raise any questions :)

chamalabey:
@yashvardhan-kukreja I have made the changes as you have requested. 
Also added global variables to store URL path prefix so that we only have to change one value when old API is deprecated.

yashvardhan-kukreja:
Thanks @chamalabey , will review them :)

managedtenants/utils/ocm.py
```diff
@@ -29,6 +29,8 @@ def retry_hook(exception):
 
 class OcmCli:
     API = "https://api.stage.openshift.com"
+    ADDON_MGMT_API_URL_PREFIX = "/api/clusters_mgmt/v1/addons"
+    ADDON_MGMT_NEW_API_URL_PREFIX = "/api/addons_mgmt/v1/addons"
```
yashvardhan-kukreja, line 33:
```suggestion
    CLUSTERS_SERVICE_ADDON_MGMT_API_URL_PREFIX = "/api/clusters_mgmt/v1/addons"
    ADDONS_SERVICE_ADDON_MGMT_API_URL_PREFIX = "/api/addons_mgmt/v1/addons"
```

wdyt

yashvardhan-kukreja, line 33:
```suggestion
    CLUSTERS_SERVICE_ADDON_MGMT_API_URL_PREFIX = "/api/clusters_mgmt/v1/addons"
    ADDONS_SERVICE_ADDON_MGMT_API_URL_PREFIX = "/api/addons_mgmt/v1/addons"
```

wdyt

managedtenants/utils/ocm.py
```diff
@@ -104,67 +106,100 @@ def token(self):
...
+            {metadata.get("id")}/versions''',
+            json=addon)
 
-    def _add_addon_version(self, imageset, metadata):
+
+    # Update Tooling to point to new addon-service API MTSRE-601
+    def add_addon_version_as(self, imageset, metadata):
+        # Create the addon first if it does not exisit
+        if self._addon_exists(metadata.get("id")) is False:
+            self.add_addon(metadata)
```
yashvardhan-kukreja, line 148:
```suggestion
            self.add_addon_as(metadata)
```

managedtenants/utils/ocm.py
```diff
@@ -104,67 +106,100 @@ def token(self):
...
 
     def add_addon_version(self, imageset, metadata):
-        if self._addon_exists(metadata.get("id")):
-            return self._add_addon_version(imageset, metadata)
-        # Create the addon first
-        self.add_addon(metadata)
-        return self._add_addon_version(imageset, metadata)
+        # Create the addon first if it does not exisit
+        if self._addon_exists(metadata.get("id")) is False:
+            self.add_addon(metadata)
```
yashvardhan-kukreja, line 136:
```suggestion
            self._add_addon_version(metadata)
```

yashvardhan-kukreja, line 136:
```suggestion
            return self._add_addon_version(imageset, metadata)
```

managedtenants/utils/ocm.py
```diff
@@ -104,67 +106,100 @@ def token(self):
...
+        addon = self._addon_from_imageset(imageset, metadata)
+        return self._post(
+            f'''{self.ADDON_MGMT_API_URL_PREFIX}/
+            {metadata.get("id")}/versions''',
+            json=addon)
 
-    def _add_addon_version(self, imageset, metadata):
+
+    # Update Tooling to point to new addon-service API MTSRE-601
+    def add_addon_version_as(self, imageset, metadata):
```
yashvardhan-kukreja, line 145:
```suggestion
    def _add_addon_version_as(self, imageset, metadata):
```

managedtenants/utils/ocm.py
```diff
@@ -104,67 +106,100 @@ def token(self):
...
+        if self._addon_exists(metadata.get("id")) is False:
+            self.add_addon(metadata)
+        addon = self._addon_from_imageset(imageset, metadata)
+        return self._post(
+            f'''{self.ADDON_MGMT_API_URL_PREFIX}/
+            {metadata.get("id")}/versions''',
+            json=addon)
 
-    def _add_addon_version(self, imageset, metadata):
+
```
yashvardhan-kukreja, line 143: Let's restore the `_add_addon_version` method so that we can still use it for CS-related methods like it was being used previously

managedtenants/utils/ocm.py
```diff
@@ -104,67 +106,100 @@ def token(self):
...
-        self.add_addon(metadata)
-        return self._add_addon_version(imageset, metadata)
+        # Create the addon first if it does not exisit
+        if self._addon_exists(metadata.get("id")) is False:
+            self.add_addon(metadata)
+        addon = self._addon_from_imageset(imageset, metadata)
+        return self._post(
+            f'''{self.ADDON_MGMT_API_URL_PREFIX}/
+            {metadata.get("id")}/versions''',
+            json=addon)
```
yashvardhan-kukreja, line 141:
```suggestion
        return self._add_addon_version(imageset, metadata)
```

managedtenants/utils/ocm.py
```diff
@@ -104,67 +106,100 @@ def token(self):
...
         except OCMAPIError:
             return False
 
     def add_addon_version(self, imageset, metadata):
-        if self._addon_exists(metadata.get("id")):
-            return self._add_addon_version(imageset, metadata)
-        # Create the addon first
-        self.add_addon(metadata)
-        return self._add_addon_version(imageset, metadata)
+        # Create the addon first if it does not exisit
```
yashvardhan-kukreja, line 134:
```suggestion
        # Create the addon first if it does not exist
```

managedtenants/utils/ocm.py
```diff
@@ -307,6 +342,17 @@ def upsert_addon(self, metadata):
             raise exception
         return addon
 
+    # Add/update addon to addon service(MTSRE-601)
+    def addon_service_upsert_addon(self, metadata):
```
yashvardhan-kukreja, line 346:
```suggestion
    def addons_service_upsert_addon(self, metadata):
```

managedtenants/utils/ocm.py
```diff
@@ -317,6 +363,16 @@ def upsert_addon_version(self, imageset, metadata):
             raise exception
         return addon
 
+    # Post addon version data to versions endpoint or addon service(MTSRE-601)
+    def addon_servic_upsert_addon_version(self, imageset, metadata):
```
yashvardhan-kukreja, line 367:
```suggestion
    def addons_service_upsert_addon_version(self, imageset, metadata):
```

managedtenants/utils/ocm.py
chamalabey: I have shorten the Variable name as follows 
```
    CS_ADDON_MGMT_API_URL_PREFIX = "/api/clusters_mgmt/v1/addons"
    AS_ADDON_MGMT_API_URL_PREFIX = "/api/addons_mgmt/v1/addons"
```

yashvardhan-kukreja: perfect, thanks!

yashvardhan-kukreja (approved):
Thanks @chamalabey , lgtm :)

PR #281, opened 2022-11-07
[MTSRE-515] feat: introduce monitoringStack API to the metadata

yashvardhan-kukreja:

# py-mtcli

## Description

introduces the support for the `monitoringStack` field in the metadata schema for specifying the config for the monitoring stack CR to be utilised by the addon-operator in runtime. 

## Checklist

### DIFF
--- a/managedtenants/schemas/metadata.schema.yaml
+++ b/managedtenants/schemas/metadata.schema.yaml
@@ -130,6 +130,39 @@ properties:
     description: "Annotations to be applied to all objects created in the SelectorSyncSet."
   monitoring:
     type: object
+    description: "[DEPRECATED] Configuration parameters to be injected in the ServiceMonitor used for federation. The target prometheus server found by matchLabels needs to serve service-ca signed TLS traffic (https://docs.openshift.com/container-platform/4.6/security/certificate_types_descriptions/service-ca-certificates.html), and it needs to be runing inside the monitoring.namespace, with the service name 'prometheus'."
+    additionalProperties: false
+    required:
+      - portName
+      - namespace
+      - matchNames
+      - matchLabels
+    properties:
+      portName:
+        type: string
+        pattern: ^[A-Za-z0-9][A-Za-z0-9-]{0,60}[A-Za-z0-9]$
+        description: "The name of the service port fronting the prometheus server."
+      namespace:
+        type: string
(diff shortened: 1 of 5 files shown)


managedtenants/schemas/metadata.schema.yaml
```diff
@@ -129,6 +129,39 @@ properties:
...
+        description: "List of series names to federate from the prometheus server."
+      matchLabels:
+        type: object
+        items:
+          type: string
+          format: printable
+          pattern: ^[a-zA-Z_][a-zA-Z0-9_]*$
+        description: "List of labels used to discover the prometheus server(s) to be federated."
+        minItems: 1
+  metricsFederation:
```
ashishmax31, line 164: Why is this renamed to `metricsFederation`?

tests/sss/test_federated_metrics.py
```diff
@@ -8,6 +8,94 @@
 from managedtenants.core.addon_manager import AddonManager
 from managedtenants.core.addons_loader.sss import Sss
 
+@given(data=hypothesis_strategies.data())
+def test_metrics_federation_v1(data):
```
ashishmax31, line 12: This was existing functionality right? You just added tests for them right?

managedtenants/schemas/metadata.schema.yaml
yashvardhan-kukreja: previously metrics-federation config was provided under the `monitoring` section. And after realising that we need a new `monitoringStack` field, it was realised that `monitoring` and `monitoringStack` fields both in the addon metadata is pretty bad naming. Hence, the idea of renaming/deprecating the `monitoring` field to `metricsFederation` was formed.

That's why, if you notice, the description of the `monitoring` field begins with `[DEPRECATED]`

PS: the `monitoringStack` is totally unrelated to these `monitoring` and `metricsFederation` fields by the way

tests/sss/test_federated_metrics.py
yashvardhan-kukreja: These tests were added to ensure that the selectorsyncsets are rendered properly. Just added them for the sake of better coverage in favor of backwards compatibility. Although it's pretty much not possible for us to move back to SelectorSyncSet, no harm in having some extra tests ;)

ashishmax31 (approved):

PR #295, opened 2022-11-24
Remove adding array indices as ids for additionalCatalogSources

nikhil-thomas:
Remove call to `index_dict` in 2 occasions

- ie, treat `additionalCatalogSources` as a normal value eg: `addon[mapped_key] = val` (ref: https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/utils/ocm.py#L446)

Signed-off-by: Nikhil Thomas <nikthoma@redhat.com>

# py-mtcli

## Description

- modified X
- modified Y
- modified Z

## Checklist


nikhil-thomas:
`id` in additionalCatalog sources is required as per openshift api spec.

(ref: additionalCatalogSource in https://api.openshift.com/)

/close

PR #330, opened 2023-03-16
fix: update regex pattern to allow hyphen(-)

Ankit152:
## Description

Allow regex pattern to accept hyphen for `matchLabels` field in `monitoring/metricsFederation`.


yashvardhan-kukreja:
can be closed in favor of #331

PR #331, opened 2023-03-17
fix: object schemas for labels and annotations

yashvardhan-kukreja:
# py-mtcli

## Description

The existing way to defining `object` in our schemas did not enforce the regex validation it intended to enforce.

For example,

A schema
```json
{
           "matchLabels": {
           	"type": "object",
            "items": {
            	"type": "string",
                "pattern": "^[a-zA-Z][a-zA-Z0-9_-]*$"
            }
           }
}
```

still continued to allow the following input

```json
{
    	"matchLabels": {
        	"wrong with : $": "foo with spaces x"
        }
}
```

which is unexpected.

This PR fixes that by making use of the `patternProperties` which allow us to enforce regex restriction on the keys of on objects as well as its values.


Ankit152:
LGTM

PR #332, opened 2023-03-17
chore: sanitize payload before sending to AS

ashishmax31:


managedtenants/utils/ocm.py
```diff
@@ -130,6 +130,19 @@ def _sanitize_addon_metadata_for_addons_service(self, addon, metadata):
 
         return addon
 
+    def _sanitize_addon_imageset_for_addons_service(self, addon, imageset, metadata):
+        mapped_key = self.IMAGESET_KEYS["addOnParameters"]
+        # This will remove order attribute in addon.
+        if imageset.get("addOnParameters") is None:
+            addon[mapped_key] = metadata.get("addOnParameters", [])
+        else:
+            addon[mapped_key] = imageset.get("addOnParameters", [])
```
yashvardhan-kukreja, line 139: but we just wanna remove the `order` attribute in each element under addOnParameters, not the addonParameteres array itself :/

ashishmax31: no we are not removing them, we are just removing the extra "items" attribute and using the stuff from the metadata/imageset. (Which don't have an order attribute)

ashishmax31: This is how its already being done: https://github.com/mt-sre/managed-tenants-cli/blob/f85b9f5ac7b216a2d976a6f203afce353a348c31/managedtenants/utils/ocm.py#L126

yashvardhan-kukreja: ohh nvm, I went crazy for a bit and misread the `.get(.., <default-value>)` syntax as `addon[mapped_key] = []` 

(feel like Friday)
:joysob:

yashvardhan-kukreja (approved):
/lgtm

PR #353, opened 2023-05-03
Added deps pins for urllib3 and requests

npecka:
# py-mtcli

## Description

Due to docker-py dependencies we needed to add pins for urllib3 and requests until fix is in place


npecka:
Fixed with PyPI pulling 2.30 from requests

PR #373, opened 2024-04-04
feat(MTSRE-1839): make base index images configurable

Ajpantuso:


Nanyte25 (approved):
/lgtm

managedtenants/bundles/index_builder.py
```diff
@@ -12,12 +12,15 @@
...
 class IndexBuilder:
+    # pylint: disable=too-many-arguments
     def __init__(
         self,
         docker_api,
         dry_run=False,
         debug=False,
         build_with="digest",
+        # pylint: disable=line-too-long
+        base_image="quay.io/mtsre/opm-ubi@sha256:fcaa5e9cb99bf6df239e63b3412bce972937aeee9b7c74862dec07b048b7c569",  # noqa: E501
```
apahim, line 23: still a bit weird to have the default this deep into the code... maybe have the default set in the `bundles_parser.add_argument` and here you always asume there's a value?

Ajpantuso: Fixed, didn't want to create a breaking scenario for callers of this `__init__` method, but not a big deal. The only one referenced in this code base is properly updated.

apahim (approved):
lgtm

